PR #9, opened 2020-04-09
traj.py

aring1988:
Traj.py hysplit trajectory reader added


bbakernoaa (changes requested):
Can you rename this.  Just thinking if someone adds a different trajectory model in the future.  Something like hytraj.py

bbakernoaa:
@aring1988 please make sure to add this to the __init__.py in monetio/models add it to the import in monetio/__init__.py

bbakernoaa (approved):

bbakernoaa:
looking good @aring1988 Merging

PR #19, opened 2020-06-17
Update the GOES 16 and 17 readers

bbakernoaa:
fix: This adds a proper reader for the NOAA GOES 16 and GOES 17 L1 and L2 products.  It also allows for reading of the GOES data both locally or from the amazon S3 server.

### DIFF
--- a/monetio/__init__.py
+++ b/monetio/__init__.py
@@ -29,6 +29,7 @@ icartt = profile.icartt
 tolnet = profile.tolnet
 
 # sat
+goes = sat.goes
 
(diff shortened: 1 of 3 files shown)


bbakernoaa:
The datasets can be easily be used like so

```python
import monetio as mio
mio.goes.open_dataset(date='2020-05-10 22:00', satellite='16', product='ABI-L2-AODF')
```

bbakernoaa:
@amcz @aring1988 Have either of you had time to look at this?

Issue #31, opened 2021-05-13
hysplit get_latlongrid function bug


amcz:
grid points in the HYSPLIT output cdump file represent the center of the sampling area.
This was coded correctly in the makegrid method in the ModelBin class. However, the get_latlongrid function was
shifting the location of the grid by half a grid point. Since get_latlongrid is utilized by the combine_dataset function
to align the files, this resulted in grid from combine_dataset being shifted half a grid point with respect to the when using
the open_dataset method.

amcz:
Resolved.

PR #32, opened 2021-05-13
Hysplitdev

amcz:
Resolves issue #31. grid points in the HYSPLIT output cdump file represent the center of the sampling area.
This was coded correctly in the makegrid method in the ModelBin class. However, the get_latlongrid function was
shifting the location of the grid by half a grid point. Since get_latlongrid is utilized by the combine_dataset function
to align the files, this resulted in grid from combine_dataset being shifted half a grid point with respect to the when using
the open_dataset method.

### DIFF
--- a/monetio/models/hysplit.py
+++ b/monetio/models/hysplit.py
@@ -28,6 +28,11 @@ Classes
 --------
 ModelBin
 
+Change log
+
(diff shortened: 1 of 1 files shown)


amcz:
Resolved.

PR #39, opened 2021-07-29
Development: Addition of HRRR-Smoke and RAQMS readers

mebruckner:
Added readers developed to bring in HRRR-Smoke and RAQMS model data.

### DIFF
--- a/monetio/__init__.py
+++ b/monetio/__init__.py
@@ -24,6 +24,7 @@ ncep_grib = models.ncep_grib
 hysplit = models.hysplit
 hytraj = models.hytraj
 pardump = models.pardump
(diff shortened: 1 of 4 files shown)


zmoon:
@bbakernoaa does this look good to merge?

bbakernoaa:
@mebruckner can we include the command in which you converted the grib2 data to netcdf.  There are many ways that this can be done.....was it with eccodes or wgrib2? Have you looked into cfgrib? maybe we can just read the hrrr_smoke data difrectly with xarray

```python
xr.open_dataset('hrrr_smoke.grib',engine='cfgrib')
```

Do you perhaps have a small example file for each of the hrrr_smoke and raqms files?

PR #45, opened 2022-02-03
AERONET fixes

### LINKED ISSUE #14: aeronet and sda data
currently aeronet can read the aerosol AOD level 1->2 data and the total AOD level 1->2 but is having a problem with the SDA products.

### LINKED ISSUE #15: add lunar aeronet flag
AERONET started adding data using a lunar retrieval to get data at night time.  This option needs to be added to the url generation.  need to add a lunar=False flag to be able to retrieve this data if wanted

zmoon:
Resolves #14 #15

Additionally:
* store info lines from top of AERONET files in `DataFrame.attrs`
  - use this info to handle some failure situations a bit better
* add pytest-xdist to speed up testing a bit


monetio/obs/aeronet.py
```diff
@@ -95,6 +96,9 @@ def add_data(
         if both are specified.
     daily : bool
         Load daily averaged data.
+    lunar : bool
+        Load provisional lunar "Direct Moon" data instead of the default "Direct Sun".
+        Only for non-version products.
```
zmoon:
```suggestion
        Only for non-inversion products.
```

PR #48, opened 2022-02-09
RRFS CMAQ

zmoon:


zmoon:
@drnimbusrain do you want this to be merged?

zmoon:
maybe we don't need to since M-M has developed one as well that should eventually get in here. https://github.com/NOAA-CSL/MELODIES-MONET/blob/develop/melodies_monet/new_monetio/rrfs_cmaq.py Or we could merge this one now and later merge the two. @bbakernoaa ?

bbakernoaa:
@zmoon Lets use the one in melodies Monet

zmoon:
#52 added the current MM RRFS-CMAQ reader

Issue #49, opened 2022-02-22
bugfix to hysplit.py


amcz:
EMC was finding failure in parse_hdata6and7 when computing datetime objects pdate1 and pdate2.  The elements of the numpy dtype need to be accessed like hdata6["omonth"][0]. They were being accessed like hdata6["omonth"]. The latter appears to work in some setups but not all.

amcz:
This issue has been fixed.

PR #51, opened 2022-02-25
Hysplitdev

amcz:
Fixes issue #49 EMC was finding failure in parse_hdata6and7 when computing datetime objects pdate1 and pdate2. The elements of the numpy dtype need to be accessed like hdata6["omonth"][0]. They were being accessed like hdata6["omonth"]. The latter appears to work in some setups but not all.

Also changes to formatting and linting of code.


zmoon:
@amcz looks like there are just a few little things Black wants to be changed. Otherwise looks good to me!

PR #52, opened 2022-03-02
Add in MM Reader Scripts for WRF-Chem, CMAQ, RRFS-CMAQ, and CESM-FV

rschwant:
This updates adds the Melodies Monet Reader Scripts for WRF-Chem, CMAQ, and RRFS-CMAQ into monetio

### DIFF
--- a/monetio/models/__init__.py
+++ b/monetio/models/__init__.py
@@ -1,4 +1,4 @@
-from . import camx, cmaq, fv3chem, hysplit, hytraj, ncep_grib, pardump, prepchem 
+from . import camx, cmaq, fv3chem, hysplit, hytraj, ncep_grib, pardump, prepchem
 from . import _cmaq_mm, _wrfchem_mm, _rrfs_cmaq_mm
 
 __all__ = ["cmaq", "hysplit", "camx", "fv3chem", "ncep_grib", "prepchem", "hytraj", "pardump",
--- a/monetio/models/__init__.py
+++ b/monetio/models/__init__.py
@@ -1,7 +1,31 @@
-from . import camx, cmaq, fv3chem, hysplit, hytraj, ncep_grib, pardump, prepchem
-from . import _cmaq_mm, _wrfchem_mm, _rrfs_cmaq_mm
+from . import (
+    _cesm_fv_mm,
+    _cmaq_mm,
+    _rrfs_cmaq_mm,
+    _wrfchem_mm,
+    camx,
+    cmaq,
+    fv3chem,
+    hysplit,
+    hytraj,
+    ncep_grib,
+    pardump,
+    prepchem,
+)
 
-__all__ = ["cmaq", "hysplit", "camx", "fv3chem", "ncep_grib", "prepchem", "hytraj", "pardump",
-          "_cmaq_mm","_wrfchem_mm","_rrfs_cmaq_mm"]
+__all__ = [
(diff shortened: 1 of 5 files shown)


rschwant:
Oh I just realized I point to monetio in these scripts and I should update how this is done now that these scripts are in the monetio package. Let me fix this. This may be why these checks are failing.

zmoon:
@rschwant the failure I saw was because you are trying to use something from monet. Is it possible to avoid that? I'm not sure but it looks like it is unused?

zmoon:
Also would you be able to run the pre-commit hooks on your branch? (<https://pre-commit.com/#quick-start>) and fix the few things flake8 would like? I can do it if you are too busy.

zmoon:
> I just realized I point to monetio in these scripts

I don't think it will cause problems to leave them as absolute imports instead of relative

rschwant:
I'm still working on this. I need to make some updates to the CESM reader. The plots don't look correct. Probably just some relabeling and converting the longitude from 0 to 360 to -180 to 180. I'll try to get this done tonight.

rschwant:
I updated the CESM reader to be more similar in structure to the other readers, but it did not correct the contour plot problem. I realized that we need to set roll_dateline=True in the quick_contourf plot in plotting routines in MELODIES MONET to fix this. I can add this to the list of things to update, but it can be next week.

zmoon:
@rschwant test failing because of `pyresample` not being in the test conda env. Could you move imports of `pyresample` inside function(s) that need it instead of module scope (there are examples of this elsewhere in monetio)? and for wrf-python (`wrf`) as well? currently these are not monetio required deps

and I can do if you are too busy

rschwant:
I ran the pre-commit @zmoon , which is a very cool tool by the way. Does this exist in MELODIES MONET too? 

The other checks look like they are failing because I added a function from pyresample

Can we not use pyresample in monetio? 

Can I call MONET in these scripts? This seemed to not be allowed either. I can also comment this line out for now, but eventually I want these reader scripts to put the model output in a consistent format including longitude on -180 to 180 instead of 0 to 360.

rschwant:
Oh sorry just realized you emailed me the exact same thing. Yes I will update this.

zmoon:
> I ran the pre-commit @zmoon , which is a very cool tool by the way. Does this exist in MELODIES MONET too?

not yet, I have been meaning to ask you if you were interested! I can set it up for sure.

rschwant:
> > I ran the pre-commit @zmoon , which is a very cool tool by the way. Does this exist in MELODIES MONET too?
> 
> not yet, I have been meaning to ask you if you were interested! I can set it up for sure.

Yes, definitely interested, but let's do it after we get a stable version working, so maybe later March early April.

zmoon:
> Can I call MONET in these scripts? This seemed to not be allowed either. I can also comment this line out for now, but eventually I want these reader scripts to put the model output in a consistent format including longitude on -180 to 180 instead of 0 to 360.

yeah, maybe we can address this later. Ideally `monetio` should not require `monet`; the idea is more that a monet workflow would most likely require monetio, but monetio is independent. It sounds like maybe this functionality you want could be moved into monetio

zmoon (approved):
We're passing now and you have tested that these work, so let's merge this. I'll try to add some tests for these readers when we make them "official"

PR #53, opened 2022-03-04
Move CESM SE reader file to monetio

Duseong:
I moved the CESM SE reader file to monetio.


zmoon (commented):
Linters are complaining but otherwise looks good to go

Duseong:
Sorry I forgot to run pre-commit hooks, so I ran these, and I think it's OK now.

zmoon (approved):

PR #56, opened 2022-03-08
Updated CMAQ open_mfdataset.

drnimbusrain:
Recently tried to use monetio to read in multiple CMAQ files, and there was an issue when combining multiple files along a specified dimension:

ValueError: When combine='by_coords', passing a value for `concat_dim` has no effect. To manually combine along a specific dimension you should instead specify combine='nested' along with a value for `concat_dim`.

### DIFF
--- a/monetio/models/cmaq.py
+++ b/monetio/models/cmaq.py
@@ -115,7 +115,7 @@ def open_mfdataset(
     """
 
     # open the dataset using xarray
-    dset = xr.open_mfdataset(fname, concat_dim="TSTEP", **kwargs)
+    dset = xr.open_mfdataset(fname, combine='nested', concat_dim="TSTEP", **kwargs)
 
     # add lazy diagnostic variables
     dset = add_lazy_pm25(dset)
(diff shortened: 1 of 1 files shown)


zmoon:
Yeah, we also had to update the MELODIES MONET CESM reader because of this.

This became an error in xarray v0.20.0: https://xarray.pydata.org/en/stable/whats-new.html#id32

monetio/models/cmaq.py
```diff
@@ -115,7 +115,7 @@ def open_mfdataset(
     """
 
     # open the dataset using xarray
-    dset = xr.open_mfdataset(fname, concat_dim="TSTEP", **kwargs)
+    dset = xr.open_mfdataset(fname, combine='nested', concat_dim="TSTEP", **kwargs)
```
zmoon:
```suggestion
    dset = xr.open_mfdataset(fname, combine="nested", concat_dim="TSTEP", **kwargs)
```
linting should pass with double quotes here

zmoon (changes requested):

monetio/models/cmaq.py
zmoon: I know you marked this resolved, but we do need this for linting to pass, so I'm just going to commit it.

zmoon (approved):

PR #70, opened 2022-06-27
Fixes for MM data proc examples

zmoon:
https://github.com/NOAA-CSL/MELODIES-MONET/pull/88

* [x] AERONET
* [x] AQS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,11 @@ jobs:
       - name: Test with pytest
         run: pytest -n auto -v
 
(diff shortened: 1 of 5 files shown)


rschwant:
@zmoon, great thanks for fixing this so quickly. This looks like it will fix all of the problems quite well. And it looks like we just need to update MONETIO and no changes are needed to the MELODIES-MONET reader scripts, correct?

zmoon:
> no changes are needed to the MELODIES-MONET reader scripts, correct?

Yes I think they should work without additional modifications now. I will update monetio.

PR #71, opened 2022-07-25
Develop satellite

dwfncar:


zmoon:
Hi @dwfncar, I resolved the merge conflict but am not able to push. Maybe if you "Allow edits by maintainers" in the PR somewhere?

(Note that `develop` and `stable` in the NCAR fork are a bit behind.)

dwfncar:
Don't see many options in the PR itself.
I've added you as a maintainer to the NCAR/monetio fork.
Can you try to merge the conflict after accepting the repo invite?
thanks,
David

zmoon:
@dwfncar it is supposed to be [like this](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) but maybe it is different since the NCAR fork is under an org not a personal account

dwfncar:
I don't have that option.  I think it needs to be a user-owned fork instead
of institutional.

zmoon (changes requested):
Initial comments:

* can you modify the hdfio test so that the test file is deleted after it is complete?
* `pyhdf` is currently not a required dep and I would prefer to keep it that way. Possible resolutions include:
  - modify the `hdfio` module so that the module-level import is not needed (within functions, you could use the `_import_required` helper seen in geoms reader)
* if `hdfio` is going to be the only module, we could probably remove the `hdf` directory
* for the new readers, is it possible to get tiny test files so that we can test them?
  - if not, please prepend `_` and append `mm` to the module names

monetio/models/raqms.py
```diff
@@ -0,0 +1,111 @@
...
+            f = xr.open_dataset(names[0], drop_variables=["theta"])
+            f = _fix_grid(f)
+            f = _fix_time(f)
+        else:
+            raise ValueError
+    except ValueError:
+        print(
+            """File format not recognized. Note that files should be
+                preprocessed to netcdf."""
+        )
```
zmoon:
```suggestion
        print(
            "File format not recognized. Note that files should be "
            "preprocessed to netcdf."
        )
```
otherwise unintended indentation

monetio/models/raqms.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+            f = _fix_grid(f)
+            f = _fix_time(f)
+        else:
+            raise ValueError
+    except ValueError:
+        print(
+            """File format not recognized. Note that files should be in netcdf
+                format. Do not mix and match file types."""
+        )
```
zmoon:
```suggestion
        print(
            "File format not recognized. Note that files should be "
            "in netCDF format. "
            "Do not mix and match file types."
        )
```
otherwise unintended indentation

monetio/sat/__init__.py
```diff
@@ -1,12 +1,12 @@
...
-    modis_l2,
     modis_ornl,
     nesdis_edr_viirs,
     nesdis_eps_viirs,
     nesdis_frp,
-    omps_limb,
-    omps_nadir,
+    _modis_l2_mm,
+    _omps_limb_mm,
+    _omps_nadir_mm,
```
zmoon: I suspect the import sorter will want these at the top

tests/test_hdfio.py
```diff
@@ -59,3 +61,5 @@ def tearDown(self):
 
     test_suite = unittest.TestSuite([unittest.TestLoader().loadTestsFromTestCase(test_hdf)])
     unittest.TextTestRunner(verbosity=2).run(test_suite)
+    os.remove('test.hdf')
```
zmoon: Unfortunately this doesn't help when running with pytest. It looks like it can't go within tearDown because setUp/tearDown is done for every test, so the file is gone after the first test.

zmoon: The updated test set unfortunately doesn't work with the parallel test runner (try with `pytest -n 2 -v -k hdf`)

zmoon: I tried something in 6bbfe80

monetio/models/raqms.py
dwfncar: trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check docstring is first.................................................Passed
check yaml...........................................(no files to check)Skipped
pyupgrade................................................................Passed
isort....................................................................Passed
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted monetio/models/raqms.py

All done! ✨ 🍰 ✨
1 file reformatted.

black formatter combined some of the lines

tests/test_hdfio.py
zmoon: @dwfncar are you satisfied with that resolution?

dwfncar: Yes

monetio/models/__init__.py
```diff
@@ -12,6 +12,7 @@
     ncep_grib,
     pardump,
     prepchem,
+    raqms,
```
zmoon, line 15: Can we get a small test file for RAQMS? Otherwise might want to give it the `_` / `_mm` treatment too

dwfncar: Let me check with Maggie

zmoon: @mebruckner I added some tests in a765270. I also made some editorial changes to the reader. Can you check that everything looks fine to you?

The main change was that I adjusted the condition for raising `ValueError` to be more strict, and then also changed it to actually raise instead of just print.

zmoon:
Is fv3raqms reader supposed to be in this PR?

mebruckner:
No, I was keeping that one in house as I don't think FV3-RAQMS is really public. To my knowledge we aren't currently producing forecasts and are just doing a bunch of experiments. I accidentally pushed my copy of the driver to the Melodies-Monet develop_satellite branch without the fv3ramqs reader syntax commented out.
________________________________

monetio/models/__init__.py
zmoon: @mebruckner can you confirm updated code here works for your larger set of files?

zmoon:
`omps_limb` removed on request of @mebruckner

zmoon (approved):
I didn't look at the details of the two modules we are leaving "hidden". If at some point we want to make them public in monetio I will take a closer look.

PR #75, opened 2022-08-15
Remove module-level Dask progress bar registration

### LINKED ISSUE #73: Remove module-level Dask progress bar registration
Currently present in four places.

Related to https://github.com/NOAA-CSL/MELODIES-MONET/issues/95

zmoon:
Resolves #73 

The issue is that the way the monetio is currently set up, all that code gets run when you import `monetio`, so you get progress bars even if you don't want them (e.g. for job logs).

### DIFF
--- a/monetio/obs/aqs.py
+++ b/monetio/obs/aqs.py
@@ -3,15 +3,10 @@ import os
 import warnings
 
 import pandas as pd
-from dask.diagnostics import ProgressBar
 
 from .epa_util import read_monitor_file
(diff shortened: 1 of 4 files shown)


zmoon:
If you do want to see progress bars (e.g. in a notebook), just do
```python
from dask.diagnostics import ProgressBar

ProgressBar().register()
```

[more info](https://docs.dask.org/en/stable/diagnostics-local.html)

PR #79, opened 2022-08-18
Be more strict about RRFS-CMAQ reader PM2.5 merge

### LINKED ISSUE #78: RRFS-CMAQ reader should error when PM2.5 ds coords don't match
https://github.com/NOAA-CSL/MELODIES-MONET/issues/94#issuecomment-1219797459

Currently this reader supports adding separate PM2.5 files, as it currently isn't in the default output. If the coords don't match, the merge
https://github.com/noaa-oar-arl/monetio/blob/9bb2a013fe6e8907b80a206041dfced14cf73cd7/monetio/models/_rrfs_cmaq_mm.py#L165
leads to skipped plots or plots with confusing results in MM, as mentioned in the thread above.

It would be better to do some checks before the merge and raise a nice error message if necessary.

cc @rschwant

zmoon:
Fixes #78

* [x] test of this exception handling

### DIFF
--- a/monetio/util.py
+++ b/monetio/util.py
@@ -396,8 +396,17 @@ def _try_merge_exact(left, right, *, right_name=None):
         # Try to print more debug info
         import re
 
-        regex = r"not equal along these coordinates \(dimensions\): '(?P<name>[a-zA-Z0-9_]*)'"
-        m = re.search(regex, str(e))
+        name = r"(?P<name>[a-zA-Z0-9_]*)"
+
+        m = None
+        for regex in [
+            rf"not equal along these coordinates \(dimensions\): '{name}'",
+            # Older message, used up to at least 0.21.1:
+            rf"indexes along dimension '{name}' are not equal",
+        ]:
+            m = re.search(regex, str(e))
+            if m is not None:
+                break
         if m is None:
             warnings.warn(
                 f"Unexpected Exception message (expected to match {regex!r}): {e}", stacklevel=2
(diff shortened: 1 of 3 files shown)


zmoon:
The Python 3.6 run here was getting xarray 0.18.2, which has a different message in this situation. But now accounted for.

zmoon:
@rschwant if you can take a quick look at some point that would be great

monetio/util.py
```diff
@@ -376,3 +376,51 @@ def _import_required(mod_name: str):
         raise RuntimeError(
             f"importing required module '{mod_name}' failed. {_install_message(mod_name)}"
         ) from e
+
+
+def _try_merge_exact(left, right, *, right_name=None):
+    """For two ``xr.Dataset``s, try ``left.merge(right, compat="equals", join="exact")``.
+    If it fails, print informative debugging messages and re-raise.
+    Otherwise, return the result.
+    """
```
zmoon, line 385:
```suggestion
    `right_name` is used for context in the error messages if provided.
    """
```

rschwant:
This looks good to me. I like the detail of the error messages, so that it is really clear when something goes wrong. Was Jianping able to fix the float issue in his scripts, so that these files can merge properly now?

zmoon:
Thanks @rschwant 

> Was Jianping able to fix the float issue in his scripts, so that these files can merge properly now?

I haven't gotten an update on that. I had the same issue with some RRFS-CMAQ output from Patrick, and I did some pre-processing using `ds = ds.merge(ds_pm, join="exact", compat="override")`, which afaik at least checks that the index shapes match.

PR #81, opened 2022-09-15
Short fix to _cesm_se_mm reader for cases in which only surface variables are present in the var_list.

dwfncar:
Append lev to var_list if not present, otherwise the rename of lev to z will fail in cases in which only surface variables are being read.


zmoon (approved):
Seems only fair, given that lat/lon get this treatment.

PR #84, opened 2022-09-19
AERONET interp for daily product

### LINKED ISSUE #82: AERONET wl interp doesn't seem to work with the daily product
Something with differences in column names.

Specifically, the daily product has columns like `'n[aod_1640nm]'`, `'n[aod_1020nm]'`, which have `aod` in the name.

zmoon:
Resolves #82 

* Fix AOD column identification so that interp works for the daily product, which has some different columns
* Warn and rename old AOD column(s) when a requested wavelength to interp to is the same as one of the existing nominal wavelength columns


zmoon:
linkcheck failure due to <https://www.ready.noaa.gov> currently being down it looks like

PR #101, opened 2023-02-14
Fix pardump reading on some platforms

### LINKED ISSUE #99: pardump not reading READY HYSPLIT PARDUMP.#### files
Hello, 
I'm attempting to use MONETIO to read the PARDUMP file generated by running HYSPLIT on the READY server. The "self.pardt" formatting in the MONETIO "pardump.open_dataset" function doesn't seem to be the same as the generated PARDUMP.#### (#### = job number) file. Here is what I'm doing:

Input: (minus path)
`import monetio as mio
import glob

folderpath = "path"
partfiles=glob.glob(folderpath+"*/PARDUMP*")
part = mio.pardump.open_dataset(partfiles[0],century=2000,verbose=1)`

Output:
`Record Header  [(28, 10020, 1, 22, 12, 24, 23, 0)]
Date  2022-12-24 23:00:00  ****  None
Adding data  [(28, 10020, 1, 22, 12, 24, 23, 0)] 2022-12-24 23:00:00
Record Header  [(7, 0, 1, 1, 8907, 20, 4, 940279797)]
Traceback (most recent call last):

  File "\\sitc2\Planning\USER\WATER\6_AIR_QUALITY\C_AQ_BKGD_RESEARCH\4_PROGRAMS\Python\Plot_Hysplit.py", line 21, in <module>
    part = mio.pardump.open_dataset(partfiles[0],century=2000,verbose=1)

  File "C:\Users\klarson\Anaconda3\lib\site-packages\monetio\models\pardump.py", line 36, in open_dataset
    pardf = pdump.read(drange=drange, century=century, verbose=verbose)

  File "C:\Users\klarson\Anaconda3\lib\site-packages\monetio\models\pardump.py", line 191, in read
    pdate = datetime.datetime(

ValueError: month must be in 1..12`

Using the Debugger, I looked at the par_frame before the padding is dropped. There seems to be an offset as the pandas array fills "lon", "ht", and "su" with the latitude, longitude, and height values (see attached screen shot). I'm not sure if this is a general issue with all PARDUMP HYSPLIT files or if this is solely a PARDUMP generated-on-the-READY-server HYSPLIT file issue.  I have the same dispersion back trajectory from READY file I was attempting to read attached as a zip file below. 

par_frame

[2022122500.zip](https://github.com/noaa-oar-arl/monetio/files/10700980/2022122500.zip)

zmoon:
Resolves #99 

In our testing, the issue seems to be that on Windows, dtype `'>l'` (l as in long) resolves to `'>i4'`, but it is supposed to be `'>i8'` for this file format. On Linux, don't have this issue.

### DIFF
--- a/monetio/models/pardump.py
+++ b/monetio/models/pardump.py
@@ -52,8 +52,8 @@ class Pardump:
         self.fname = fname
         # self.dtfmt = "%Y%m%d%H%M"
 
-        tp1 = ">f"  # big endian float.
-        tp2 = ">i"  # big endian integer.
+        tp1 = ">f4"  # big endian float.
+        tp2 = ">i4"  # big endian integer.
         tp3 = ">i8"  # big endian long integer.
 
         # header record in fortran file.
@@ -144,7 +144,7 @@ class Pardump:
             hdr["minute"] = sdate.minute
             print(hdr)
 
-            endrec = np.array([20], dtype=">i")
+            endrec = np.array([20], dtype=">i4")
 
             fpoint.write(hdr)
             fpoint.write(a)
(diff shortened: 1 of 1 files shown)


zmoon:
@amcz could you maybe confirm that results are still what you expect with this change?

amcz:
> @amcz could you maybe confirm that results are still what you expect with this change?

Yes. I have checked and confirm that on the linux platform the reader still works as expected. Thanks.

PR #105, opened 2023-04-06
Fix ISH-lite reader

### LINKED ISSUE #104: ish_lite reader broken
The current version of the `ish_lite` reader is broken. It is unable to retrieve data.

Currently there are two issues in that 1) the read_history file didn't have a the dates passed to it.  This needs to be fixed. 2) the build_url needs to ensure that only the available urls are read.  

A side issue is that none of the meta data is added back to the returned dataframe from add_data.

bbakernoaa:
This fixes multiple issues in the `ish_lite` reader #104 .  

- Adds verbose flag to print out step by step changes for debugging purposes
- fix the history reader to get the dates available correctly
- build_url now ensures that it only reads data available by directly looking at each year url to gather all the individual site urls
- slims the data to only the target dates 
- merges the meta data from history back into the frame


monetio/obs/ish_lite.py
```diff
@@ -4,7 +4,15 @@
...
 
 def add_data(
-    dates, box=None, country=None, state=None, site=None, resample=True, window="H", n_procs=1
+    dates,
+    box=None,
+    country=None,
+    state=None,
+    site=None,
+    resample=True,
+    window="H",
```
zmoon: Currently these are unused. Should we add?

environment-dev.yml
```diff
@@ -16,6 +16,7 @@ dependencies:
   #
   # optional
   - joblib
+  - lxml
```
zmoon, line 19: pandas `read_html` needed this (or bs4)

bbakernoaa: sorry didn't know that was there.  it was just in my environment

monetio/obs/ish_lite.py
bbakernoaa: it isn't a bad idea.  I thought about it but didn't try.  ish_lite by default is hourly where ish could be variable (15 minutes, 30minutes, 60minutes, etc)

zmoon: I added in a69353c

zmoon: I fixed it so the resample works now, but since it is already hourly, shouldn't it be false by default?

environment-dev.yml
zmoon: No prob, I think it's fine to leave things this way, since the pandas message is reasonably clear.

monetio/obs/ish_lite.py
bbakernoaa: Good idea

monetio/obs/ish_lite.py
```diff
@@ -127,31 +138,37 @@ def read_data_frame(self, file_object):
         frame = pd.DataFrame.from_records(frame_as_array)
         df = self._clean(frame)
```
zmoon: This is not defined. Maybe we could just remove `read_data_frame` and `get_url_file_objs` ?

bbakernoaa: Yes we can remove the `read_data_frame` routine.  It's a carry over from the `ish` reader.

the `get_url_file_objs` allows for downloading the data raw data. if this isn't required we can remove it

zmoon: Left the latter for now.

Issue #103, opened 2023-03-01
add unit testing for hysplit.py and hytraj.py


amcz:
Add test_hysplit.py and test_hytraj.py  to unit tests folder with some basic unit testing.

amcz:
added into hysplitdev branch

zmoon:
Now in `develop`

Issue #102, opened 2023-03-01
hytraj  would like to have function to combine several tdump files into one DataFrame


amcz:
Add function similar to the hysplit combine_dataset function which will take as input a list of trajectory output (tdump) files and
return one pandas DataFrame with information from all the files.

amcz:
Added into hysplitdev branch

zmoon:
In `develop`

PR #108, opened 2023-04-12
Hysplitdev

amcz:
merge hysplitdev back into develop. 
Addresses issues #102 adding function to hytraj to combine several tdump files into one dataframe.
Addresses issue  #103 add some unit testing for hysplit.py and hytraj.py
Small changes to hysplit.py functions 
    get_latlongrid improved exception statements
    get_latlon replace x>=180 test with x>=180+lon_tolerance
    get_thickness modified to calculate if attribute specifying vertical
    levels is bad.


zmoon:
@amcz the tests did not all pass. Generally we want everything to be passing before merging. Also I would have been happy to review a bit.

PR #109, opened 2023-04-20
Fix ISH for empty sites

zmoon:
Discovered some bugs related to empty sites while setting up ISH support for MM.
* for partially empty (i.e. one or more but not all sites missing all data for that period), bytes conversion to string would fail, now the method just returns the input df if empty
* for fully empty, the resample groupby would fail, now skipped if empty

### DIFF
--- a/monetio/obs/ish.py
+++ b/monetio/obs/ish.py
@@ -238,8 +238,8 @@ class ISH:
 
         The constructed 'station_id' column is a combination of the USAF and WBAN columns.
         This is done since USAF and WBAN alone are not unique in the history file.
-        For example, USAF 725244 and 722158 appear twice, as do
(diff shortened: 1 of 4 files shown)


zmoon:
Note EPA AQS link check ~~was briefly failing~~ is failing intermittently

`
(           index: line   73) broken    https://www.epa.gov/aqs/ - HTTPSConnectionPool(host='origin-awswest-www.epa.gov', port=443): Max retries exceeded with url: /aqs (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f04b087e880>: Failed to establish a new connection: [Errno 110] Connection timed out'))
`

usually it succeeds if I just run it a second time though, weird

bbakernoaa (approved):

PR #110, opened 2023-04-20
Fix hysplitdev tests

zmoon:
#108 was merged with tests and lints failing


zmoon:
linkcheck failing since <https://matplotlib.org/tutorials/colors/colormaps.html> currently down

also a few tests are failing with pandas v2, so upper-bounding for now

Issue #113, opened 2023-05-06
Faster ISD


zmoon:
Currently it is pretty slow. Youhua reported that it took somewhat over an hour to get all US sites for a period with one unique year.

Andrew Lambert of NRL said that with his code he was able to do something similar (vis data only though?) loading [from Amazon S3](https://aws.amazon.com/marketplace/pp/prodview-tmvgq72wrnzpq) in 3 minutes or so. So there is a lot of room for improvement.

---

I did some initial benchmarking on my desktop, picked a site year with larger file size: site 01010099999, year 2020, 748K. Loading with `pd.read_fwf`[^1], specifying `widths` and `dtype`s (also using `header=None`, `compression="gzip"`). 3 options for this fixed-width (FW) file:
* `s3://noaa-isd-pds/data/2020/010100-99999-2020.gz` ~ 0.7 s
* <https://noaa-isd-pds.s3.amazonaws.com/data/2020/010100-99999-2020.gz> ~ 1 s
* <https://www1.ncdc.noaa.gov/pub/data/noaa/2020/010100-99999-2020.gz> ~ 1.4 s

:point_up: So there is a factor of 2 speed gain in this case by using the S3 URL instead. But note not all the processing that monetio does is included.

Loading this site-year similarly
```python
ish.add_data(pd.date_range("2020/01/01", "2021/01/01", freq="D")[:-1], site="01010099999", resample=False)
```
(though more processing done, e.g. 99999 -> NaN) with monetio currently takes ~ 15 s. Probably by leveraging `pd.read_fwf` etc. like above instead of the current method, we can speed the processing up.

Also note that loading the CSV (`s3://noaa-global-hourly-pds/2020/01010099999.csv`) instead of the FW text seems not that much slower. Will have to compare with all the processing included etc.

Also note that <https://www.ncei.noaa.gov/data/global-hourly/archive/> has compiled files (all sites, presumably) by year. These are just `.tar.gz`s of all the CSV or FW files though.

[^1]: [`pd.read_fwf`](https://pandas.pydata.org/docs/reference/api/pandas.read_fwf.html) args are not all documented in its docstring, but they are mostly the same as [`pd.read_csv`](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html), with the differences noted [here](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#files-with-fixed-width-columns)

---

cc: @ytangnoaa @bbakernoaa

[...]

PR #116, opened 2023-05-26
Fix OpenAQ reader

zmoon:
* local time calc
  - don't quite understand *why* it wasn't working properly (giving only a few unique UTC offsets), but looks more reasonable now (several unique UTC offsets and correct e.g. most common is +8 UTC which mostly corresponds to sites in China)
* convert to ppmv
  - there was code to do so but it wasn't being used
  - updated and documented the conversion factors and added factors for CH4 and NO which [can be](https://docs.openaq.org/docs/parameters) in OpenAQ data

* [x] at least one test (maybe find a way to set things to load just one file for testing)

### DIFF
--- a/monetio/obs/openaq.py
+++ b/monetio/obs/openaq.py
@@ -199,10 +199,13 @@ class OPENAQ:
                 o3="o3_ppm",
                 no2="no2_ppm",
                 so2="so2_ppm",
+                ch4="ch4_ppm",
+                no="no_ppm",
                 bc="bc_umg3",
                 pm25="pm25_ugm3",
                 pm10="pm10_ugm3",
             ),
             axis=1,
+            errors="ignore",
         )
         return w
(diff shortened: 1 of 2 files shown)


zmoon:
Some initial timings suggest we can gain speed by using `s3` URL scheme instead of current HTTPS:
```
In [8]: %timeit -n 10 pd.read_json("https://openaq-fetches.s3.amazonaws.com/realtime/2013-11-26/2013-11-26.ndjson",
   ...:  lines=True)
311 ms ± 37.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit -n 10 pd.read_json("s3://openaq-fetches/realtime/2013-11-26/2013-11-26.ndjson", lines=True)
159 ms ± 13.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
similar to #113

PR #126, opened 2023-07-11
Deal with AirNow UTC offsets that are 0 but shouldn't be

### LINKED ISSUE #86: Some AirNow sites with incorrect UTC offset of 0
There are 20-ish TX sites and 1 WA site that have UTC offset set to 0 erroneously (exact numbers/sites depend on time period). I reported this to AirNow, and they said they will fix, but monetio could still do a quick check for this based on lon and just drop those rows.

and/or we could use timezonefinder (if installed / as an optional dependency) to compute the values.

This was noticed by @colin-harkins

zmoon:
* [x] add option to drop (or fix?) those rows
* [x] document those options

Resolves #86

### DIFF
--- a/monetio/obs/airnow.py
+++ b/monetio/obs/airnow.py
@@ -177,6 +177,10 @@ def aggregate_files(dates=dates, *, download=False, n_procs=1, daily=False, bad_
         before loading.
     n_procs : int
         For Dask.
+    bad_utcoffset : {'null', 'drop', 'fix', 'leave'}, default: 'drop'
+        How to handle bad UTC offsets
+        (i.e. rows with UTC offset 0 but abs(longitude) > 20 degrees).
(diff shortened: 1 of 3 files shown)


bbakernoaa (approved):

monetio/obs/airnow.py
```diff
@@ -275,9 +295,72 @@ def filter_bad_values(df, *, max=3000):
...
+        elif bad_utcoffset == "leave":
+            pass
+        else:
+            raise ValueError("`bad_utcoffset` must be one of: 'null', 'drop', 'fix', 'leave'")
+
     return df  # TODO: dropna here (since it is called `filter_bad_values`)?
 
 
+def get_utcoffset(lat, lon):
+    """Get UTC offset for standard time.
```
zmoon:
```suggestion
    """Get UTC offset for standard time (hours).
```

Issue #122, opened 2023-06-13
Faster OpenAQ (using openaq-fetches)


zmoon:
Some initial timings suggest we can gain speed by using `s3` URL scheme instead of current HTTPS:
```
In [8]: %timeit -n 10 pd.read_json("https://openaq-fetches.s3.amazonaws.com/realtime/2013-11-26/2013-11-26.ndjson",
   ...:  lines=True)
311 ms ± 37.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: %timeit -n 10 pd.read_json("s3://openaq-fetches/realtime/2013-11-26/2013-11-26.ndjson", lines=True)
159 ms ± 13.5 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
similar to #113

_Originally posted by @zmoon in https://github.com/noaa-oar-arl/monetio/issues/116#issuecomment-1585078076_

zmoon:
Also was experimenting at some point and found some ways to speed up the JSON parsing, datetime in particular.

PR #141, opened 2023-10-26
OpenAQ v1 fixes + initial OpenAQ v2 reader

### LINKED ISSUE #130: OpenAQ fixes
Aside from the plan to increase the speed (#122), while working on an MM CLI for @jordanschnell I found some things that should be fixed in the current OpenAQ reader that uses openaq-fetches.

* [x] returned df currently may have some times outside of the requested dates (i.e. [greater than the ub](https://github.com (/noaa-oar-arl/monetio/blob/01e14d36d7b43e559a2f817d109657c0ddfc34a4/monetio/obs/openaq.py#L123)) (in #141)
* [ ] returned df may have non-unique site IDs at given time.
  Seems like there are two things going on here:
  - lat/lon that are only slightly different, the precision of the lat/lon in the current site IDs doesn't capture. Maybe create new site IDs, e.g. country code + hash of full-precision lat/lon (in #141)
  - lat/lon are the same but different measurements in different rows (e.g. 2 rows). Should combine these into one row where possible.
* [x] typo in BC's new variable name with units (fixed in 01e14d3)
* [x] add some tests which load a reduced set of files (in #141)

zmoon:
Closes #130 #122


zmoon:
AQS unreachable again.

And for 3.8, additionally getting:
```
FAILED tests/test_openaq.py::test_openaq_first_date - TypeError: openssl_sha1() takes at most 1 argument (2 given)
FAILED tests/test_openaq.py::test_openaq_long_fmt - TypeError: openssl_sha1() takes at most 1 argument (2 given)
FAILED tests/test_openaq.py::test_openaq_2023 - TypeError: openssl_sha1() takes at most 1 argument (2 given)
```

bbakernoaa (approved):
I would just make sure we document where to receive the API key

zmoon:
Thanks @bbakernoaa . I have this automatic warning set up: https://github.com/noaa-oar-arl/monetio/blob/4490a50d7e76bcc7d2e10d33feb5ff5ef69c53dc/monetio/obs/openaq_v2.py#L60-L67 but I will also add the relevant info to the module docstring.

PR #142, opened 2023-10-26
For AirNow UTC offset test, use yesterday instead of current day

zmoon:
Addressing test failures in #141 that are unrelated to those changes.

If day just turned, AirNow might not have data yet (seemed to be what was happening).


zmoon:
AQS websites "unreachable"

PR #145, opened 2023-10-26
Docs work: mainly surface obs

zmoon:
* moving code examples to a notebook that can be run (though takes a while)
  - some fixes made in order for some examples to work, though some things still not working
* including documentation about a data source in the module docstring


zmoon:
With #260 won't need these docs/examples updates. But there may be some other useful stuff in here...

PR #154, opened 2023-12-14
Add RRFS-SD reader

### LINKED ISSUE #152: RRFS-SD Reader
It came to my attention that the RRFS-SD verification has been using the `_rrfs_cmaq_mm.py` reader.  While this works it is probably better to create a separate one that is a simpler use case than the complex chemistry of cmaq itself. 

Specially there are three variables for air composition within the files.  

`smoke` (pm25) 
`dust` (pm25) 
`coarse_pm` (PM10 coarse mode only) 

We need to aggregate these into PM25 and PM10

PM25 = smoke + dust

PM10 = smoke + dust + PM10

@jordanschnell can you confirm the variable names here

bbakernoaa:
create a separate reader for RRFS-SD.  This is based on `_rrfs_cmaq_mm` but removes many of the functions as rrfs-sd is obviously much simpler.


bbakernoaa:
this addresses #152

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
++        index (int): The index to check.
++    
++    Returns:
++        bool: True if the index is the maximum value, False otherwise.
++    """
+    if index.max():
+        return True
+    else:
+        return False
+
```
zmoon: Looks to be unused
```suggestion
```

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
+""" RRFS-CMAQ File Reader """
+import numpy as np
+import xarray as xr
+from numpy import concatenate
```
zmoon:
```suggestion
```

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+from pandas import Series
```
zmoon:
```suggestion
```

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+    -------
+    xarray.DataSet
+        RRFS-SD model dataset in standard format for use in MELODIES-MONET
+
+    """
+
+    # Read in all variables and do all calculations.
+    dset = xr.open_mfdataset(fname, concat_dim="time", combine="nested", **kwargs)
+    
+    if 'smoke' in dset.data_vars and 'dust' in dset.data_vars:
```
zmoon: Simplify
```suggestion
    if {'smoke', 'dust'} <= dset.data_vars.keys():
```

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+        dset['PM25'] = dset.smoke + dset.dust
+        dset['PM25'].attrs['long_name'] = 'Particulate Matter < 2.5 microns'
+        dset['PM25'].attrs['units'] = "ug/kg"
+    if 'smoke' in dset.data_vars and 'dust' in dset.data_vars and 'coarsepm' in dset.data_vars:
```
zmoon:
```suggestion
    if {'smoke', 'dust', 'coarsepm'} <= dset.data_vars.keys(): 
```

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+
+    
+
+    # convert "ug/kg to ug/m3"
+    for i in dset.variables:
+        if "units" in dset[i].attrs:
+            if "ug/kg" in dset[i].attrs["units"]:
+                # ug/kg -> ug/m3 using dry air density
+                dset[i] = dset[i] * dset["pres_pa_mid"] / dset["temperature_k"] / 287.05535
+                dset[i].attrs["units"] = r"$\mu g m^{-3}$"
```
zmoon, line 95: If we really want a pretty version here could use unicode, e.g. `μg/m³` or `μg m⁻³`. But maybe better to use just ASCII `ug/m3` or `ug m-3`.

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+    dset = dset.set_coords(["latitude", "longitude"])
+
+    # These sums and units are quite expensive and memory intensive,
+    # so add option to shrink dataset to just surface when needed
+    if surf_only:
+        dset = dset.isel(z=0).expand_dims("z", axis=1)
+
+    
+
+    # convert "ug/kg to ug/m3"
```
zmoon, line 89: Do we want this to be optional like Jordan's?

zmoon (commented):
Just few initial comments. I'll work on the pressure calculation once I have extracted some sample data.

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Read in all variables and do all calculations.
+    dset = xr.open_mfdataset(fname, concat_dim="time", combine="nested", **kwargs)
+    
+    if 'smoke' in dset.data_vars and 'dust' in dset.data_vars:
+        dset['PM25'] = dset.smoke + dset.dust
+        dset['PM25'].attrs['long_name'] = 'Particulate Matter < 2.5 microns'
+        dset['PM25'].attrs['units'] = "ug/kg"
+    if 'smoke' in dset.data_vars and 'dust' in dset.data_vars and 'coarsepm' in dset.data_vars: 
+        dset['PM10'] = dset.smoke + dset.dust + dset.coarsepm
```
zmoon: Might be better to use `pm10` for consistency. And currently in MM it would cause a problem for model var to have same name as obs var.

monetio/models/_rrfs_sd_mm.py
bbakernoaa: Let’s stick with ascii. It may cause issues down the pipeline if we save the raw files out. It might not be recognized as CF convention otherwise

monetio/models/_rrfs_sd_mm.py
bbakernoaa: Also let’s rename it to pm10_sd and pm25_sd

monetio/models/_rrfs_sd_mm.py
zmoon: Then I think `ug m-3` best for CF (though they would want `kg m-3` for canonical).

monetio/models/_rrfs_sd_mm.py
bbakernoaa: I dont see why

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+
+    # Read in all variables and do all calculations.
+    dset = xr.open_mfdataset(fname, concat_dim="time", combine="nested", **kwargs)
+    
+    if {'smoke', 'dust'} <= dset.data_vars.keys():
+        dset['pm25_sd'] = dset.smoke + dset.dust
+        dset['pm25_sd'].attrs['long_name'] = 'Particulate Matter < 2.5 microns'
+        dset['pm25_sd'].attrs['units'] = "ug/kg"
+    if {'smoke', 'dust', 'coarsepm'} <= dset.data_vars.keys():
```
bbakernoaa, line 38: didn't know about this syntax.  Really clean and useful.

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,166 @@
...
+
+def _calc_pressure(dset, surf_only=False):
+    """Calculate the mid-layer pressure in Pa from surface pressure
+    and ak and bk constants.
+
+    Interface pressures are calculated by:
+    phalf(k) = a(k) + surfpres * b(k)
+
+    Mid layer pressures are calculated by:
+    pfull(k) = (phalf(k+1)-phalf(k))/log(phalf(k+1)/phalf(k))
```
bbakernoaa, line 140: not sure where this formula came from

bbakernoaa: This is the formula used in the post_fv3 https://github.com/NOAA-EMC/fv3atm/blob/a82381c0b751a15e5343de5078ef836b2c444c89/io/post_fv3.F90#L4266-L4276

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,166 @@
...
+    p = dset.dp_pa.copy().load()  # Have to load into memory here so can assign levels.
+    srfpres = dset.surfpres_pa.copy().load()
+    for k in range(len(dset.z)):
+        if surf_only:
+           pres_2 = 0.0 + srfpres * 0.9978736
+           pres_1 = 0.0 + srfpres * 1.0
+        else:
+           pres_2 = dset.ak[k + 1] + srfpres * dset.bk[k + 1]
+           pres_1 = dset.ak[k] + srfpres * dset.bk[k]
+        p[:, k, :, :] = (pres_2 - pres_1) / np.log(pres_2 / pres_1)
```
zmoon, line 161: Maybe something like this
```suggestion
    surfpres = dset.surfpres_pa
    a = dset.ak
    b = dset.bk
    if surf_only:
        phalf_kp1 = 0.0 + surfpres * 0.9978736
        phalf_k = 0.0 + surfpres * 1.0
    else:
        phalf_kp1 = a.shift(z=-1) + surfpres * b.shift(z=-1)
        phalf_k = a + surfpres * b
    p = (phalf_kp1 - phalf_k) / np.log(phalf_kp1 / phalf_k)
```
going for consistency with the docstring variable names

zmoon:
@bbakernoaa I made an extraction (forecast hour 6 of the run you shared, 5 levels closest to surface, selected variables, etc.), now available at https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_model_data/rrfssd_example/rrfs-sd_dynf006.nc (104M) for testing.

We can set it up like I did [here](https://github.com/noaa-oar-arl/monetio/blob/6d2cb70a63f967d1a2b587cf2e0028e396eb0d14/tests/test_mopitt_l3.py), which also uses data stored in that location.

monetio/models/_rrfs_sd_mm.py
zmoon: Since `ak` and `bk` are dataset attrs, not variables
```python
phalf_kp1 = a[1:] + surfpres * b[1:]
```
but may need to make them `DataArray`s for the dims to match up properly and such, maybe create
```python
a_k = xr.DataArray(ds.ak[:-1], dims="z")
a_kp1 = xr.DataArray(ds.ak[1:], dims="z")
b_k = xr.DataArray(ds.bk[:-1], dims="z")
b_kp1 = xr.DataArray(ds.bk[1:], dims="z")
```

monetio/models/_rrfs_sd_mm.py
bbakernoaa: @zmoon still not sure where this formula came from

bbakernoaa: @rschwant do you know? I think that you originally did this

bbakernoaa:
> @bbakernoaa I made an extraction (forecast hour 6 of the run you shared, 5 levels closest to surface, selected variables, etc.), now available at https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_model_data/rrfssd_example/rrfs-sd_dynf006.nc (104M) for testing.
> 
> We can set it up like I did [here](https://github.com/noaa-oar-arl/monetio/blob/6d2cb70a63f967d1a2b587cf2e0028e396eb0d14/tests/test_mopitt_l3.py), which also uses data stored in that location.

maybe we can add some netcdf tricks to compress that even further using integers instead of floats and the add_offset and scale_factor netcdf attributes

monetio/models/_rrfs_sd_mm.py
rschwant: Raffaele sent it to me to describe the vertical structure in the model when I was helping with the boundary conditions. I'll forward you all the email. This is how he described it to me.

Interface pressure levels are computed using the hybrid interface formula:
p(k) = a(k) + ps * b(k)

where ps is the (actual/reference) surface pressure. These pressure levels correspond to phalf in your output dyn*.nc files, while pfull are computed as:
pfull(k) = (phalf(k+1)-phalf(k))/log(phalf(k+1)/phalf(k))

If there is an official typical way of calculating this we can use that instead too. We haven't really tested this much since we have not used the aircraft evaluation in MELDODIES MONET much in the RRFS-CMAQ model.

bbakernoaa: @rschwant I think that we should perhaps use what is in post_fv3 https://github.com/NOAA-EMC/fv3atm/blob/a82381c0b751a15e5343de5078ef836b2c444c89/io/post_fv3.F90#L4270

rschwant: Sure, there may have been less variables available at the time, so if there is a newly suggested method, let's go with that. Is there a user guide that describes the vertical structure now to confirm and reference our approach with this calculation?

bbakernoaa: I've been trying to dig through them and I haven't found where it is.  But it uses the same vertical coordinate as GEOS and what is listed there is the same.

zmoon:
> maybe we can add some netcdf tricks to compress that even further using integers instead of floats and the add_offset and scale_factor netcdf attributes

Perhaps, but I think it is better to keep format closer to the original. And I used NCO lossy compression, which already decreases the size a lot due to quantization, I don't know if additionally doing the int packing transform would make it any more compressible.

monetio/models/_rrfs_sd_mm.py
zmoon: I'll compare the two methods' results.

zmoon: There are differences but they are all less than 0.2 hPa in the column I tested.

<details>

Differences in Pa:
image

image
</details>

bbakernoaa: this is how Raffaele calculated it for UFS-Aerosols in the exglobal_aero_init function https://github.com/NOAA-EMC/global-workflow/blob/develop/ush/merge_fv3_aerosol_tile.py#L99-L101

bbakernoaa: I think that we should use that

zmoon: That is calculation of delta p, similar to `dpres` in the dataset but not the same:

<details>

image
</details>

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,166 @@
...
+    f : xarray.Dataset
+        RRFS-CMAQ model data
+
+    Returns
+    -------
+    xr.DataArray
+        Geoptential height with attributes.
+    """
+    sfc = f.surfalt_m.load()
+    dz = f.dz_m.load() * -1.0
```
zmoon, line 121: This (loading all dz) is what Maggie's run is failing on, specifically:

```
Unable to allocate 129. GiB for an array with shape (37, 65, 2961, 4881) and data type float32
```

Like the pressure calc we should be able to write this in a Dask-friendly way.

zmoon: Also like pressure calc, for the surf-only case there is a short version.

monetio/models/_rrfs_sd_mm.py
```diff
@@ -0,0 +1,166 @@
...
+    # These are negative in RRFS-CMAQ, but you resorted and are adding from the surface,
+    # so make them positive.
+    dz[:, 0, :, :] = dz[:, 0, :, :] + sfc  # Add the surface altitude to the first model level only
+    z = dz.rolling(z=len(f.z), min_periods=1).sum()
```
zmoon, line 125: I would think `.cumsum()` could be used.

PR #157, opened 2024-01-17
Add retry and timeout options for ISH reader

zmoon:
and also address a few warnings.

Default timeout is 10 (like it was set to before), but now, by default, 4 retries will be made, for a total of 5 tries. And both timeout and retries can be set from `add_data`.

cc: @jordanschnell

xref: NOAA-CSL/MELODIES-MONET#232


zmoon:
Retries shouldn't be needed after moving to AWS (#113), but we might set it up so using AWS would be optional.

zmoon:
As noted [here](https://github.com/NOAA-CSL/MELODIES-MONET/pull/236#issuecomment-1898902919), a historical case with full-length files did succeed, though took > 2 hours.

bbakernoaa (approved):

Issue #132, opened 2023-09-06
Add reader for VIIRS AOD on S3


zmoon:
> The full archive of reprocessed VIIRS aerosol optical depth (AOD) & aerosol detection (ADP) granules EDR data files are now via NOAA Open Data Distribution (NODD)!

https://noaa-jpss.s3.amazonaws.com/index.html#SNPP/VIIRS/
https://noaa-jpss.s3.amazonaws.com/index.html#NOAA20/VIIRS/

cc: @bbakernoaa

zmoon:
@bbakernoaa mentioned gridded, so maybe 
* https://noaa-jpss.s3.amazonaws.com/index.html#NOAA20/VIIRS/NOAA20_VIIRS_Aerosol_Optical_Depth_Gridded_Reprocessed/
* ...

bbakernoaa:
this may help a lot.  I think that we should adapt this to either stream or download the data. https://github.com/jpss-nodd/python-scripts/blob/main/README.md

bbakernoaa:
This is partially addressed by #164 . We should still add in something for the swath data

PR #166, opened 2024-03-21
Feature/icap mme

bbakernoaa:
This adds on top of #132 adding the ICAP_MME to monetio.  ICAP MME is the 9 member multi member ensemble system developed at NRL. It is available through https://usgodae.org/metools/ensemble/

### DIFF
--- a/monetio/models/icap_mme.py
+++ b/monetio/models/icap_mme.py
@@ -16,7 +16,7 @@ def build_urls(dates, *, filetype="MMC", data_var="dustaod550", verbose=True):
 
     Parameters
     ----------
-    dates : pd.DatetimeIndex or iterable of datetime
+    dates : datetime-like or iterable of datetime-like
         Dates to download data for.
     filetype : {'MMC', 'C4', 'MME'}, optional
         The first MME date is 2014-11-28, and the last 2022-09-11.
@@ -69,8 +69,9 @@ def check_remote_file_exists(file_url):
         return False
 
 
-def retrieve(url, fname, *, download=False, stream=True, verbose=True):
-    """Download the file at `url` to path `fname` if it doesn't exist.
+def retrieve(url, fname, *, download=False, verbose=True):
+    """Return BytesIO of the file at `url` or
+    download it to path `fname` if it doesn't exist.
 
     Parameters
     ----------
@@ -94,14 +95,14 @@ def retrieve(url, fname, *, download=False, stream=True, verbose=True):
     p = Path(fname).absolute()
 
     if not download:
-        r = requests.get(url, stream=stream)
+        r = requests.get(url, stream=True)
         r.raise_for_status()
         return BytesIO(r.content)
     else:
         if not p.is_file():
             if verbose:
                 print(f"Downloading {url} to {p.as_posix()}")
-            r = requests.get(url, stream=stream)
+            r = requests.get(url, stream=True)
             r.raise_for_status()
             with open(p, "wb") as f:
                 f.write(r.content)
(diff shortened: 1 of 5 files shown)


monetio/models/icap_mme.py
```diff
@@ -0,0 +1,189 @@
...
+        print("\n File Exists: " + fname)
+
+
+def open_dataset(date, product="MMC", data_var="modeaod550"):
+    """
+    Parameters
+    ----------
+    date : str or datetime-like
+        The date for which to open the dataset.
+        2022-10-29 to current is available.
```
zmoon: Looks like the first non-MME is 2022-09-01
and the first MME is 2014-11-28

zmoon: Also we should provide a description of what MME, C4, MMC mean

bbakernoaa: MME is the same thing as MMC but for whatever reason the changed the name from MultiModelEnsemble to MultiModelConcensus

bbakernoaa: C4 is the Consensus from the 4 models with DA

monetio/models/icap_mme.py
```diff
@@ -0,0 +1,189 @@
+import pandas as pd
+
+server = "ftp.star.nesdis.noaa.gov"
+base_dir = "/pub/smcd/VIIRS_Aerosol/npp.viirs.aerosol.data/epsaot550/"
+
```
zmoon:
```suggestion
```
Don't seem to be used here.

monetio/models/icap_mme.py
```diff
@@ -0,0 +1,189 @@
...
+        raise ValueError(f"Invalid input for 'data_var': Valid values are {valid_data_vars}.")
+
+    urls, fnames = build_urls(d, filetype=product, data_var=data_var)
+    url = urls.values[0]
+    fname = fnames.values[0]
+    print(url)
+    print(fname)
+    if check_remote_file_exists(url) is False:
+        raise ValueError(f"File does not exist on ICAP HTTPS server: {url}")
+    retrieve(url, fname)
```
zmoon: Could make downloading to disk optional. `xr.open_dataset(io.BytesIO(r.content), engine="h5netcdf")` seems to work for loading the dataset from memory without saving to disk first.

monetio/models/icap_mme.py
zmoon: Thanks. [Seems like](https://usgodae.org/ftp/outgoing/nrl/ICAP-MME/2022/202209/) 2022-09-11 is the last MME. 2022-09-01 -- 11 have both MME and MMC.

monetio/models/icap_mme.py
bbakernoaa: oh interesting... should we add this to the nesdis_eps_viirs_nrt.py too?

monetio/models/icap_mme.py
zmoon: I added the date info to the docstrings.

bbakernoaa:
@zmoon this works great on my end

monetio/models/icap_mme.py
```diff
@@ -0,0 +1,276 @@
+"""
+ICAP-MME
```
zmoon, line 2: @bbakernoaa does this all sound okay?

monetio/models/icap_mme.py
```diff
@@ -0,0 +1,276 @@
...
+facing the operational aerosol community."
+
+The ICAP multi-model ensemble (ICAP-MME) is constructed from
+the following aerosol forecast systems:
+
+- ECMWF CAMS
+- JMA MASINGAR
+- NASA GEOS-5
+- NRL NAAPS
+- NOAA NGAC
```
bbakernoaa: This was NGAC but is now GEFS-Aerosols

bbakernoaa: There should be 9 models

monetio/models/icap_mme.py
bbakernoaa: I think Peng covered this

zmoon (approved):
@bbakernoaa I'm pretty happy with this unless you have more test ideas or something

Issue #168, opened 2024-04-04
hysplit combine_dataset bugs


amcz:
combine_dataset function in hysplit.py has a couple bugs.

1. It does not properly combine datasets in which the time coordinates are not the same. In the hysplit DataArrays the time coordinate are datetime64 objects. The xarray align method does not seem to be able to align these properly. The fix is to change the time coordinate to integers before aligning and then convert back to datetime objects after the datasets are merged.

2. It also is not properly combining datasets when both the source and ensemble dimensions are greater than 1. Most use cases have focused on when we are looking at a source term ensemble with one meteorological dataset (source dimension > 1 and ens dimension=1) or a meteorological ensemble (source dimension=1 and ensemble dimension > 1). However it should also be able to produce a combined dataset when both these dimensions > 1.

amcz:
this issue has been fixed in hysplitdev branch

zmoon:
> this issue has been fixed in hysplitdev branch

Cool, shall we PR to develop branch? Is the `hysplit_experimental` supposed to be included?

amcz:
Thanks - I will take the hysplit_experimental out and then it will be ready

PR #171, opened 2024-04-17
Hysplitdev

amcz:
Fixes for issue #168


monetio/models/hysplit.py
```diff
@@ -199,7 +207,7 @@ def __init__(
 
         if readwrite == "r":
             if verbose:
-                print("reading " + filename)
+                logger.info(f"reading {filename}")
```
zmoon: This won't show unless user changes their logging config. Maybe such `verbose=True` message should stick with `print()`.

monetio/models/hysplit.py
```diff
@@ -339,12 +347,16 @@ def parse_header(self, hdata1):
            number of starting locations in file.
         """
         if len(hdata1["start_loc"]) != 1:
-            print("WARNING in ModelBin _readfile - number of starting locations " "incorrect")
-            print(hdata1["start_loc"])
+            logger.warning(
+                f"WARNING in ModelBin {self.filename} _readfile - number of starting locations incorrect"
```
zmoon: If this is a message intended for the user, suggest using `warnings.warn()` instead of the logger. In either case, the word "warning" doesn't need to be included in the message since that should already be clear.

monetio/models/hysplit.py
```diff
@@ -368,7 +380,7 @@ def parse_hdata2(self, hdata2, nstartloc, century):
                     century = 2000
                 else:
                     century = 1900
-                print("WARNING: Guessing Century for HYSPLIT concentration file", century)
+                logger.info(f"WARNING: Guessing Century for HYSPLIT concentration file {century}")
```
zmoon: Suggest using `warnings.warn()`

zmoon (commented):
Regarding the switch to using a logger for certain messages: for messages intended for a normal user using the module to load HYSPLIT output, I would suggest using `print()` or `warnings.warn()`. The latter may be preferred for situations where there's something the user could change to avoid the warning, e.g. by fixing something wrong with their data or by changing how they are using your code. `logger.info()` or `logger.debug()` can be helpful for messages that may aid in your debugging of the module or helping a user debug but are not needed in normal vebrose or non-verbose usage.

monetio/models/hysplit.py
bbakernoaa: @amcz did you want to address this?

bbakernoaa (approved):

bbakernoaa (changes requested):
do the existing tests cover these? if not we need to create a test for the new files

bbakernoaa:
also we need to run pre-commit to get some linting issues fixed 

pre-commit run --all-files

PR #173, opened 2024-05-09
pyhdf 0.11.4 is now on cf

zmoon:
and should work

https://anaconda.org/conda-forge/pyhdf/files?version=0.11.4


zmoon:
I can see the 3.10 env got pyhdf 0.11.4 in the most recent CI run.

PR #178, opened 2024-07-02
ICARTT: support non-whole-number data interval, apply scaling factor

ytangnoaa:
Fixed the bug that does not consider the ICARTT's scaling factors and  datainterval in float-point time.

### DIFF
--- a/monetio/profile/icartt.py
+++ b/monetio/profile/icartt.py
@@ -341,7 +341,7 @@ class Dataset:
         # value is set to 0. The Mid-point time is required when it is not at the
         # average of Start and Stop times. For additional information see Section
         # 2.5 below.).
-        self.dataInterval = int(self.__readline()[0])
+        self.dataInterval = float(self.__readline()[0])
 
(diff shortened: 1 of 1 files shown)


ytangnoaa:
It is a bug fix to deal with some ICARTT files

monetio/profile/icartt.py
```diff
@@ -436,7 +436,7 @@ def __nan_miss_float(self, raw):
             v = x.replace(self.VAR[i].miss, "NaN")
             if "NaN" in v:
                 v = "NaN"
-            vals.append(float(v.strip()))
+            vals.append(float(v.strip())*self.VAR[i].scale)  # multiply with scaling factor
```
zmoon:
```suggestion
            vals.append(float(v.strip()) * self.VAR[i].scale)  # multiply with scaling factor
```

zmoon: Linting should pass with this change.

zmoon (approved):
Looks good to me

PR #180, opened 2024-08-02
Replace np.NaN by np.nan

blychs:
Required for compatibility with NumPy v2.0.0
	Does not break backwards compatibility

xref: https://github.com/NOAA-CSL/MELODIES-MONET/issues/263


zmoon (approved):
Thanks @blychs

PR #181, opened 2024-08-02
fix lat and lon in _cesm_fv_reader

blychs:
Correct _cesm_fv_mm.py reader for good functionality with MELODIES-MONET

	Delete variables to avoid duplication in melodies-monet
	Add (optional) variables for vertical coordinates:
		temperature_k
		alt_msl_m_mid
		alt_agl_m_mid
		surfpres_pa
		pres_pa_mid

PS: Sorry for submitting 2 PRs on the same day, these are what I had finished and had not submitted yet. I think that they should be tested separately.

xref: https://github.com/NOAA-CSL/MELODIES-MONET/issues/261

### DIFF
--- a/monetio/models/_cesm_fv_mm.py
+++ b/monetio/models/_cesm_fv_mm.py
@@ -1,10 +1,16 @@
-""" CESM File Reader """
+"""CESM File Reader"""
+
 import xarray as xr
 from numpy import meshgrid
 
 
 def open_mfdataset(
-    fname, earth_radius=6370000, convert_to_ppb=True, var_list=["O3", "PM25"], **kwargs
+    fname,
+    earth_radius=6370000,
+    convert_to_ppb=True,
+    var_list=["O3", "PM25"],
+    surf_only=True,
+    **kwargs,
 ):
     """Method to open multiple (or single) CESM netcdf files.
        This method extends the xarray.open_mfdataset functionality
@@ -52,6 +58,31 @@ def open_mfdataset(
     #############################
     # Process the loaded data
     # extract variables of choice
+    # If vertical information is required, add it.
+    # Assume that height and geopotential height are equal
+    if not surf_only:
+        dset_load.rename(
+            {
+                "T": "temperature_k",
+                "Z3": "alt_msl_m_mid",
+                "P0": "surfpres_pa",
+                "PMID": "pres_pa_mid",
+            }
+        )
+        # Calc height agl. PHIS is in m2/s2, whereas Z3 is in already in m
+        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.81
+        dset_load["alt_agl_m_mid"].attrs = {
+            "description": "geopot height above ground level",
+            "units": "m",
+        }
+        var_list = var_list + [
+            "temperature_k",
+            "alt_msl_m_mid",
+            "alt_agl_m_mid",
+            "surfpres_pa",
+            "pres_pa_mid",
+        ]
+
     dset = dset_load.get(var_list)
     # rename altitude dimension to z for monet use
     # also rename lon to x and lat to y
@@ -70,6 +101,9 @@ def open_mfdataset(
 
     # re-order so surface is associated with the first vertical index
     dset = dset.sortby("z", ascending=False)
+
+    # Get rid of variables lat and lon to avoid future conflicts
+    dset = dset.drop_vars(["lat", "lon"])
 
     #############################
 
--- a/monetio/models/_cesm_fv_mm.py
+++ b/monetio/models/_cesm_fv_mm.py
@@ -101,7 +101,7 @@ def open_mfdataset(
     # re-order so surface is associated with the first vertical index
     dset = dset.sortby("z", ascending=False)
 
-    # Get rid of variables lat and lon to avoid future conflicts
+    # Get rid of original 1-D lat and lon to avoid future conflicts
     dset = dset.drop_vars(["lat", "lon"])
 
     #############################
--- a/monetio/models/_cesm_fv_mm.py
+++ b/monetio/models/_cesm_fv_mm.py
@@ -92,8 +92,8 @@ def open_mfdataset(
     lon = wrap_longitudes(dset["lon"])
     lat = dset["lat"]
     lons, lats = meshgrid(lon, lat)
-    dset.coords["longitude"] = (("y", "x"), lons)
-    dset.coords["latitude"] = (("y", "x"), lats)
+    dset["longitude"] = (("y", "x"), lons)
+    dset["latitude"] = (("y", "x"), lats)
 
     # Set longitude and latitude to be only coordinates
     dset = dset.reset_coords()
--- a/monetio/models/_cesm_fv_mm.py
+++ b/monetio/models/_cesm_fv_mm.py
@@ -60,14 +60,25 @@ def open_mfdataset(
     # extract variables of choice
     # If vertical information is required, add it.
     if not surf_only:
-        dset_load.rename(
-            {
-                "T": "temperature_k",
-                "Z3": "alt_msl_m_mid",
-                "P0": "surfpres_pa",
-                "PMID": "pres_pa_mid",
-            }
-        )
+        if "PMID" not in ds_load.keys():
+            presvars = ["P0", "PS", "hyam", "hybm"]
+            if not all(pvar in list(ds_load.keys()) for pvar in presvars):
+                raise KeyError(
+                    "The model does not have the variables to calculate"
+                    + "the pressure. This can be done either with PMID or with"
+                    + "P0, PS, hyam and hybm.\n"
+                    + "If the vertical coordinate is not needed, set surface_only=True"
+                )
+            else:
+                ds_load["PMID"] = _calc_pressure(ds_load)
+        # dset_load.rename(
+        #     {
+        #         "T": "temperature_k",
+        #         "Z3": "alt_msl_m_mid",
+        #         "P0": "surfpres_pa",
+        #         "PMID": "pres_pa_mid",
+        #     }
+        # )
         # Calc height agl. PHIS is in m2/s2, whereas Z3 is in already in m
         dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.80665
         dset_load["alt_agl_m_mid"].attrs = {
@@ -149,3 +160,36 @@ def _ensure_mfdataset_filenames(fname):
     if len(netcdfs) >= 1:
         netcdf = True
     return names, netcdf
+
+
+def _calc_pressure(ds):
+    """Calculates midlayer pressure using P0, PS, hyam, hybm
+    Parameters
+    ----------
+    dset: xr.Dataset
+    Returns
+    -------
+    xr.DataArray
+    """
+    vert = dset["hyam"].lev.values
+    time = dset["PS"].time.values
+    lat = dset["PS"].lat.values
+    lon = dset["PS"].lon.values
+    n_vert = len(vert)
+    n_time = len(time)
+    n_lat = len(lat)
+    n_lon = len(lon)
+
+    pressure = np.zeros((n_time, n_vert, n_lat, n_lon))
+
+    for nlev in range(n_vert):
+        pressure[:, nlev, :, :] = (
+            dset["hyam"][nlev].values * dset["P0"].values
+            + dset["hybm"][nlev].values * dset["PS"].values
+        )
+    P = xr.DataArray(
+        data=pressure,
+        dims=["time", "lev", "lat", "lon"],
+        coords={"time": time, "lev": lev, "lat": lat, "lon": lon},
+        attrs={"description": "Mid layer pressure", "units": "Pa"},
+    )
--- a/monetio/models/_cesm_fv_mm.py
+++ b/monetio/models/_cesm_fv_mm.py
@@ -69,7 +69,7 @@ def open_mfdataset(
             }
         )
         # Calc height agl. PHIS is in m2/s2, whereas Z3 is in already in m
-        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.81
+        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.80665
         dset_load["alt_agl_m_mid"].attrs = {
             "description": "geopotential height above ground level",
             "units": "m",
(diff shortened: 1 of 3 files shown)


monetio/models/_cesm_fv_mm.py
```diff
@@ -52,6 +58,31 @@ def open_mfdataset(
...
+                "T": "temperature_k",
+                "Z3": "alt_msl_m_mid",
+                "P0": "surfpres_pa",
+                "PMID": "pres_pa_mid",
+            }
+        )
+        # Calc height agl. PHIS is in m2/s2, whereas Z3 is in already in m
+        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.81
+        dset_load["alt_agl_m_mid"].attrs = {
+            "description": "geopot height above ground level",
```
zmoon:
```suggestion
            "description": "geopotential height above ground level",
```
might as well spell it all out

monetio/models/_cesm_fv_mm.py
```diff
@@ -52,6 +58,31 @@ def open_mfdataset(
     #############################
     # Process the loaded data
     # extract variables of choice
+    # If vertical information is required, add it.
+    # Assume that height and geopotential height are equal
```
zmoon:
```suggestion
```
Not sure if this comment is needed, since the calculation is correct for geopotential height (which Z3 is in), and the fact that it is *geopotential* height is in the long name.

monetio/models/_cesm_fv_mm.py
```diff
@@ -52,6 +58,31 @@ def open_mfdataset(
...
+        dset_load.rename(
+            {
+                "T": "temperature_k",
+                "Z3": "alt_msl_m_mid",
+                "P0": "surfpres_pa",
+                "PMID": "pres_pa_mid",
+            }
+        )
+        # Calc height agl. PHIS is in m2/s2, whereas Z3 is in already in m
+        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.81
```
zmoon:
```suggestion
        dset_load["alt_agl_m_mid"] = dset_load["alt_msl_m_mid"] - dset_load["PHIS"] / 9.80665
```

monetio/models/_cesm_fv_mm.py
```diff
@@ -71,6 +102,9 @@ def open_mfdataset(
     # re-order so surface is associated with the first vertical index
     dset = dset.sortby("z", ascending=False)
 
+    # Get rid of variables lat and lon to avoid future conflicts
```
zmoon:
```suggestion
    # Get rid of original 1-D lat and lon to avoid future conflicts
```

zmoon (commented):
Thanks @blychs . Can you also address https://github.com/NOAA-CSL/MELODIES-MONET/issues/261#issuecomment-2214879221

zmoon:
Also maybe we can set up a test using https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_model_data/cesmfv_example/CAM_chem_merra2_FCSD_1deg_QFED_world_201909-01-09_small_sfc.nc . @blychs do you want to give that a try?

blychs:
Done with the .coords issue you commented. I already tested it with camchem.ipynb and it works fine.

@zmoon , I am happy to add the test. I guess it should be based on test_raqms.py ?
I am unaware of what dp_pa (in test_raqms) should be, my current code is not adding that for the vertical coordinate. Is it a pressure delta? Is it needed? It is not in the wrf-chem reader, which is the one I based the vertical coordinate from.

zmoon:
> based on test_raqms.py ?

I would base it on the TROPOMI L2 one, since that has code to download a file from the CSL website that you can adapt. You can also look at the RAQMS one for testing inspiration, but the setup for the RAQMS tests is simpler since it uses a file that is part of the repository.

And yes I believe dp_pa is the layer thickness in Pa. But I don't think it's required to have it.

blychs:
Hi Zach.
The file that you suggested only has surface information. I can create tests for that, but if there is a file with vertical information available from CSL, we could do a more thorough, 3D test. Otherwise, I could probably provide one. Let me know if that's needed

zmoon:
> could probably provide one

Do you have one that is 200 MB or less? Or that you could make so[^1]? The "small" one at https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_model_data/cesmfv_example/ is still big enough that I would be hesitant to have it automatically downloaded.

[^1]: By time selection, spatially subsetting (horiz and/or vert), removing variables, applying NCO lossy compression, etc.

blychs:
The easiest way is to subsample CSL's example to have less, which could do the trick. However, looking at that file, I realize that there are a couple of things I could do better:
a) I am assuming that PMID is there, which is not always the case. I will add the calculation of the midlayer pressure using the surface, P0, hyam and hyab if PMID is not there.
b) I assumed that the potential height Z3 is there, which is not always the case. I will add an optional calculation using the hydrostatic equation if Z3 is not there (CAM-chem FV is hydrostatic anyway). I could raise a warning to let the user know, just so as to future-proof the code in case a non-hydrostatic FV core is developed, but I think that the future is actually moving away from FV rather that trying to change it.

blychs:
@zmoon I added a test for the surface for now. However, it is my first time working with pytest, and I am not really sure if I know what I'm doing. Feel free to strongly criticize it.
Other than that, the _cesm_fv.py reader should be ready, with Z3 and PMID calculations if missing.

tests/test_cesm_fv.py
```diff
@@ -0,0 +1,71 @@
...
+    assert "latitude" in ds.coords
+    assert "longitude" in ds.coords
+    assert np.all(ds.latitude.values[0, :] == ds.latitude.values[0,0])
+    assert np.all(ds.longitude.values[:, 0] == ds.longitude.values[0,0])
+    assert tuple(ds["O3"].dims) == ('time', 'z', 'y', 'x')
+    assert ds['O3'].attrs['units'] == 'ppbv'
+
+
+def test_open_mfdataset(test_file_path):
+    ds = _cesm_fv.open_dataset(TEST_FP)
```
zmoon:
```suggestion
    ds = open_mfdataset(test_file_path)
    _test_ds(ds)
```

tests/test_cesm_fv.py
```diff
@@ -0,0 +1,71 @@
+import shutil
+import warnings
+from pathlib import Path
+
+import pandas as pd
+import pytest
+from filelock import FileLock
+
+from monetio.sat._tropomi_l2_no2_mm import open_dataset
```
zmoon:
```suggestion
from monetio.models._cesm_fv_mm import open_mfdataset
```

tests/test_cesm_fv.py
```diff
@@ -0,0 +1,71 @@
...
+    if worker_id == "master":
+        # Not executing with multiple workers;
+        # let pytest's fixture caching do its job
+        return retrieve_test_file()
+
+    # Get the temp directory shared by all workers
+    root_tmp_dir = tmp_path_factory.getbasetemp().parent
+
+    # Copy to the shared test location
+    p_test = root_tmp_dir / "cesm_fv_tst.nc"
```
zmoon:
```suggestion
    p_test = root_tmp_dir / "cesm_fv_test.nc"
```

monetio/models/_cesm_fv_mm.py
```diff
@@ -116,3 +157,98 @@ def _ensure_mfdataset_filenames(fname):
     if len(netcdfs) >= 1:
         netcdf = True
     return names, netcdf
+
+
+def _calc_pressure(dset):
+    """Calculates midlayer pressure using P0, PS, hyam, hybm
+    Parameters
```
zmoon, line 174: The headings need a blank line above them.

zmoon (commented):
This is a good start.

You can check locally if this is working using pytest. Assuming you have the necessary testing dependencies (see `environment-dev.yml`), you can run
```
pytest -k cesm_fv
```
at the repository root to run your test. You can add the `--pdb` flag to have it enter the Python debugger upon error. This can be helpful when developing the tests (and the library code too).

Also, to get the automatic linting/formatting working, install pre-commit with `conda install -c conda-forge pre-commit` or `pip install pre-commit` and then run `pre-commit install`.

monetio/models/_cesm_fv_mm.py
```diff
@@ -1,10 +1,17 @@
-""" CESM File Reader """
+"""CESM File Reader"""
+
 import xarray as xr
 from numpy import meshgrid
+import warning
```
zmoon:
```suggestion
import warnings
```

monetio/models/_cesm_fv_mm.py
```diff
@@ -116,3 +157,98 @@ def _ensure_mfdataset_filenames(fname):
...
+    lon = dset["PS"].lon.values
+    n_vert = len(vert)
+    n_time = len(time)
+    n_lat = len(lat)
+    n_lon = len(lon)
+
+    pressure = np.zeros((n_time, n_vert, n_lat, n_lon))
+
+    if "P0" not in dset.keys():
+        warning.warn("P0 not in netcdf keys, assuming 100_000 Pa")
```
zmoon:
```suggestion
        warnings.warn("P0 not in netcdf keys, assuming 100_000 Pa")
```

tests/test_cesm_fv.py
```diff
@@ -0,0 +1,71 @@
...
+            stream=True,
+        )
+        r.raise_for_status()
+        with open(p, "wb") as f:
+            f.write(r.content)
+
+    return p
+
+
+@pytest.fixture(scope="session")
```
zmoon:
```suggestion
@pytest.fixture(scope="module")
```
Changed this in the others in #183

monetio/models/_cesm_fv_mm.py
zmoon: This hasn't been addressed yet. We can mark it resolved after it's been addressed.

zmoon:
Looks like the test env needs `pyresample`. You can add it between `pyhdf` and `requests` in `environment-dev.yml`.

blychs:
I'm sorry for these new bugs. I should be able to have all of that fixed today.

blychs:
@zmoon I think that it should be ready now

tests/test_cesm_fv.py
```diff
@@ -29,7 +30,7 @@ def retrieve_test_file():
         with open(p, "wb") as f:
             f.write(r.content)
 
-    return p
+    return str(p)
```
zmoon: Converting to str shouldn't be necessary. Did you run into an error?

zmoon: Oh right the path manipulation stuff. I updated some of the other similar ones to accept `Path`. Perhaps you could do that for this one too...

zmoon:
Nice job getting pre-commit set up. codespell, however, doesn't like `pres`, thinks it a misspelling of `press`. Solution is to change the variable name or copy offending lines into the `.codespell-exclude` file.

tests/test_cesm_fv.py
zmoon: Or first for the quick fix, undo this but apply `str()` where you are passing `test_file_path` in `test_open_mfdataset`.

blychs: I did, because the reader has a function called _ensure_mfdataset_filenames() that checks if the file is a string and, if not, attempts to directly sort it. I did not change that (although I'm not sure we really need that, since xarray's default error handling is quite clear if the format is not right IMHO). 
There are two alternative solutions:
- If you also think that the check is not really needed, I can remove it, and converting into a string wouldn't be necessary.
- If you'd rather keep the check as a safeguard, I could add the possibility of checking for a PosixPath. The only reason I did not go that way is that it implies importing another module, which seemed like unnecessary overhead.

blychs:
I see no reason not to call it press.

tests/test_cesm_fv.py
blychs: > Oh right the path manipulation stuff. I updated some of the other similar ones to accept `Path`. Perhaps you could do that for this one too...

Sorry, my computer didn't update that you already mentioned this. I'll check your fixes tomorrow and update this that way.

tests/test_cesm_fv.py
```diff
@@ -0,0 +1,73 @@
...
+            shutil.copy(p, p_test)
+            return p_test
+
+
+def _test_ds(ds):
+    assert set(ds.dims) == {"time", "x", "y", "z"}
+
+    # Test coordinates
+    assert "lat" not in ds.data_vars
+    assert "lon" not in ds.data_vars
```
zmoon:
```suggestion
    assert "lat" not in ds.variables
    assert "lon" not in ds.variables
```
to check not in coords either

blychs:
@zmoon Is there still anything missing/ bug?

tests/test_cesm_fv.py
zmoon: I think this is now ok for now, we can always add `Path` support later.

zmoon:
@blychs there is one unresolved convo you could address. Also, I know you were interested in testing the vertical stuff, were you able to come up with a new test file to use for that?

blychs:
@zmoon I have not, I could get a simple one by just removing timesteps from the run you mentioned. Would that be acceptable?

monetio/models/_cesm_fv_mm.py
zmoon: Please add a blank line above "Parameters" and above "Returns". And same in your new hydrostatic height function.

zmoon:
Sure that sounds fine with me.

blychs:
@zmoon I created the file already (sorry for the delay, it's in Casper already) but realized that the original file does not have the temperature... which makes calculating the hydrostatic height impossible.
If you have any file with the Temperature, I could recreate that. Otherwise, let me know and I will try to get it from someone at NCAR.

zmoon:
> If you have any file with the Temperature

I do not, sorry. If you don't find one easily, we add that test another time.

blychs:
> > If you have any file with the Temperature
> 
> I do not, sorry. If you don't find one easily, we add that test another time.

I'd say we go for that. I can add a warning if surface_only=False, with something like "The 3D capabilities of the cesm_fv reader are still experimental and have not been properly tested. User discretion is adviced."
We currently have an approved version that doesn't work with MELODIES-MONET, so merging this would still be an improvement. As soon as I get a proper test_case, I'll add it in and the warning can be removed.

monetio/models/_cesm_fv_mm.py
```diff
@@ -29,9 +38,14 @@ def open_mfdataset(
     Returns
     -------
     xarray.DataSet
+    """
 
+    if not surf_only:
+        warnings.warn(
+            "3D data processing is still experimental in CESM-Fv (CAM-Chem), "
```
zmoon:
```suggestion
    xarray.Dataset
    """

    if not surf_only:
        warnings.warn(
            "3D data processing is still experimental in CESM-FV (CAM-Chem), "
```

zmoon (approved):
> As soon as I get a proper test_case, I'll add it in and the warning can be removed.

Sounds good, thanks.

PR #183, opened 2024-08-06
Reduce warnings

zmoon:
Address some warnings and reduce noise in CI pytest logs a bit (some warnings raised within the tests for info/debug are not needed in CI).


zmoon:
I think it may be better to change those messages to use [pytest-print](https://github.com/pytest-dev/pytest-print) instead, not be warnings at all.

zmoon:
Note that the `printer` prints don't currently show up when using xdist, regardless of verbosity setting.

PR #184, opened 2024-08-12
TROPOMI updates including pressure calculation 

zmoon:
Based on https://github.com/NCAR/monetio/pull/30/commits/98405e9c6634433ca6eb919813860e32dca523fd

cc: @dwfncar @mlirenzhenmayi @rrbuchholz

### DIFF
--- a/monetio/models/_wrfchem_mm.py
+++ b/monetio/models/_wrfchem_mm.py
@@ -88,6 +88,13 @@ def open_mfdataset(
         var_list.append("PSFC")
         # need to calculate surface pressure and dp and optionally dz here.
 
+        # Additional defaults for satellite analysis
+        var_list.append("PH")
+        var_list.append("PHB")
+        var_list.append("PB")
+        var_list.append("P")
+        var_list.append("T")
+
     var_wrf_list = []
     for var in var_list:
         if var == "pres":  # Insert special versions.
--- a/monetio/models/_wrfchem_mm.py
+++ b/monetio/models/_wrfchem_mm.py
@@ -89,11 +89,7 @@ def open_mfdataset(
         # need to calculate surface pressure and dp and optionally dz here.
 
         # Additional defaults for satellite analysis
-        var_list.append("PH")
-        var_list.append("PHB")
-        var_list.append("PB")
-        var_list.append("P")
-        var_list.append("T")
+        var_list.append("zstag")
 
     var_wrf_list = []
     for var in var_list:
@@ -101,11 +97,7 @@ def open_mfdataset(
             var_wrf = getvar(
                 wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="Pa"
             )
-        elif var == "height":
-            var_wrf = getvar(
-                wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="m"
-            )
-        elif var == "height_agl":
+        elif var in {"height", "height_agl", "zstag"}:
             var_wrf = getvar(
                 wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="m"
             )
@@ -114,6 +106,13 @@ def open_mfdataset(
         var_wrf_list.append(var_wrf)
 
     dset = xr.merge(var_wrf_list)
+
+    if not surf_only_nc:
+        # Compute layer thickness
+        dset["dz"] = (
+            dset["zstag"].diff("bottom_top_stag").swap_dims({"bottom_top_stag": "bottom_top"})
+        )
+        dset["dz"] = dset["dz"].assign_attrs({"units": "m", "long_name": "layer thickness"})
 
     # Add global attributes needed
     a_truelat1 = extract_global_attrs(wrflist[0], "TRUELAT1")
@@ -210,6 +209,7 @@ def open_mfdataset(
                 "temp": "temperature_k",
                 "height": "alt_msl_m_mid",
                 "height_agl": "alt_agl_m_mid",
+                "dz": "dz_m",
                 "PSFC": "surfpres_pa",
                 "pressure": "pres_pa_mid",
             }
--- a/monetio/sat/_tropomi_l2_no2_mm.py
+++ b/monetio/sat/_tropomi_l2_no2_mm.py
@@ -61,10 +61,12 @@ def _open_one_dataset(fname, variable_dict):
         if dct_ is None:
             dct_ = variable_dict.get(varname_, {})
         group_name = dct_.get("group", default_group)
+        if isinstance(group_name, list):
+            group_name = group_name[0]
         return _get_values(dso[group_name][varname_], dct_)
 
     for varname, dct in variable_dict.items():
-        print(varname)
+        print(f"- {varname}")
         if dct is None:
             dct = {}
 
@@ -116,8 +118,15 @@ def _open_one_dataset(fname, variable_dict):
             var = dso[group_name][varname]
             values = _get_values(var, dct)
 
+            if values.ndim == 2:
+                dims = ("y", "x")
+            elif values.ndim == 3:
+                dims = ("y", "x", "z")
+            else:
+                raise ValueError(f"unexpected ndim ({varname}): {values.ndim}")
+
             ds[varname] = (
-                ("y", "x"),
+                dims,
                 values,
                 {"long_name": var.long_name, "units": var.units},
             )
@@ -125,6 +134,7 @@ def _open_one_dataset(fname, variable_dict):
             if "quality_flag_min" in dct:
                 ds.attrs["quality_flag"] = varname
                 ds.attrs["quality_thresh_min"] = dct["quality_flag_min"]
+                ds.attrs["var_applied"] = dct.get("var_applied", [])
 
     dso.close()
 
@@ -141,19 +151,20 @@ def _get_values(var, dct):
     if fv is not None and not np.ma.is_masked(values):
         values[values == fv] = np.nan
 
-    if "fillvalue" in dct:
-        fillvalue = dct["fillvalue"]
-        values[values == fillvalue] = np.nan
+    fv2 = dct.get("fillvalue")
+    if fv2 is not None and not np.ma.is_masked(values):
+        values[values == fv2] = np.nan
 
-    if "scale" in dct:
-        values *= dct["scale"]
+    scale = dct.get("scale")
+    if scale is not None:
+        values *= float(scale)
 
-    if "minimum" in dct:
-        minimum = dct["minimum"]
+    minimum = dct.get("minimum")
+    if minimum is not None:
         values[values < minimum] = np.nan
 
-    if "maximum" in dct:
-        maximum = dct["maximum"]
+    maximum = dct.get("maximum")
+    if maximum is not None:
         values[values > maximum] = np.nan
 
     return values
@@ -167,15 +178,14 @@ def apply_quality_flag(ds):
     ds : xarray.Dataset
     """
     if "quality_flag" in ds.attrs:
-        quality_flag = ds[ds.attrs["quality_flag"]]
+        quality_flag = ds[ds.attrs["quality_flag"]]  # quality flag variable (float)
         quality_thresh_min = ds.attrs["quality_thresh_min"]
 
-        # Apply the quality thresh minimum to all variables in ds
-        for varname in ds:
-            if varname != ds.attrs["quality_flag"]:
-                logging.debug(varname)
-                values = ds[varname].values
-                values[quality_flag <= quality_thresh_min] = np.nan
+        # Apply the quality thresh minimum to selected variables
+        for varname in ds.attrs["var_applied"]:
+            logging.debug(f"applying quality flag to {varname}")
+            values = ds[varname].values
+            values[quality_flag <= quality_thresh_min] = np.nan
 
 
 def open_dataset(fnames, variable_dict, debug=False):
@@ -226,12 +236,10 @@ def open_dataset(fnames, variable_dict, debug=False):
             envvar = subpath.replace("$", "")
             envval = os.getenv(envvar)
             if envval is None:
-                print("environment variable not defined: " + subpath)
-                exit(1)
+                raise RuntimeError(f"environment variable {envvar!r} not defined: " + subpath)
             else:
                 fnames = fnames.replace(subpath, envval)
 
-    print(fnames)
     files = sorted(glob(fnames))
     granules = OrderedDict()
     for file in files:
--- a/monetio/sat/_tropomi_l2_no2_mm.py
+++ b/monetio/sat/_tropomi_l2_no2_mm.py
@@ -1,4 +1,10 @@
-"""Read TROPOMI L2 NO2 data."""
+"""Read TROPOMI L2 NO2 data.
+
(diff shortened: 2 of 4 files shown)


monetio/models/_wrfchem_mm.py
```diff
@@ -88,6 +88,13 @@ def open_mfdataset(
         var_list.append("PSFC")
         # need to calculate surface pressure and dp and optionally dz here.
 
+        # Additional defaults for satellite analysis
+        var_list.append("PH")
+        var_list.append("PHB")
+        var_list.append("PB")
+        var_list.append("P")
```
zmoon: None of these are the same as "pres"?

monetio/models/_wrfchem_mm.py
```diff
@@ -88,6 +88,13 @@ def open_mfdataset(
...
+        var_list.append("T")
```
zmoon: Is this ("T") the same as "tk"?

monetio/models/_wrfchem_mm.py
blychs: If this is setup for a WRF grid, then PH and PHB are the perturbation geopotential height and base geopotential height respectively (i. e., the height is (PH + PHB) / 9.8). P and PB are the perturbation and base pressures (pressure should be P + PB).

I haven't checked this reader in detail (nor will I be able to do so soon), but hopefully this helps.
Cheers,
Pablo

monetio/models/_wrfchem_mm.py
blychs: Unless this was pre - processed, it shouldn't be. In WRF, the variable T is the perturbation theta, with a base theta of 300 K (the fact that it's called T is admittedly very confusing).
To get the temperature in K, you need to do:
theta = 300 + ds["T"]
tk = theta * (pressure_in_pa / 100_000) ** (0.286), where 0.286 = R/Cp

As with the previous comment, I am not able to check this reader in detail now, but hopefully this helps.
Cheers
Pablo

zmoon:
Thanks @blychs for the comments.

blychs:
Hi @zmoon , I was wondering on how exactly are you planning on adding PH and PHB to the _wrfchem_mm reader. I need equivalent variables for the _camx_mm reader, and was just calling them "layer_height_agl" and "topo_height" as part of solving Issue #189 , but would rather make the names consistent with whatever you make. I also need to choose on those names for the TEMPO MM utility. (That is a bit different from just adding PH and PHB, since PH and PHB include the topographic height as bottom_top_stagg=0)
We would need to update every reader if we want TROPOMI and TEMPO to work with all of them, and having consistency would be useful.
It shouldn't be too hard to make everything consistent to whatever choice you make.

zmoon:
> how exactly are you planning on adding PH and PHB to the _wrfchem_mm reader

@blychs I was just following what Meng had done, which was just add them to the var list. So I was assuming that means they are included in the model output or can be automatically derived via wrf-python `getvar()`.

> I need equivalent variables for the _camx_mm reader, and was just calling them "layer_height_agl" and "topo_height"

The reader [already has](https://github.com/noaa-oar-arl/monetio/blob/2d4dc68cedde785d44a2623b92c5f127af8c8038/monetio/models/_wrfchem_mm.py#L205) height AGL and above MSL. I don't think the height of the topography comes out though, at least by default.

If @mlirenzhenmayi was only using PH, PHB to compute height as (PH+PHB)/g0, maybe we don't actually need them? And similar for P, PB.

blychs:
@zmoon I thought that was mid layer height, whereas PH and PHB are the geopotential at the layer interface. This is quite useful, because then (PH + PHB)[:, 1:, :, :] - (PH + PHB)[:, -1, :, :]) / g0 is the layer thickness, which is needed for calculating the partial column. If this is not the case, please let me know, because my _camx_mm reader assumes it is, and I would have to change that (that is, I calculated the height at mid layer for that variable).
(PH + PHB) / 9.81 at bottom_top_stagg=0 should be pretty much the topography, if I'm not wrong.

I don't think we need P, PB though. P + PB is just the pressure.

zmoon:
@blychs I believe you are correct about [the calculation](https://github.com/NCAR/wrf-python/blob/5c45adfcbc1e06c5d80a911f6a80b2d22f8ebdc5/src/wrf/g_geoht.py#L96). But if it's really the layer height thickness is needed, maybe we should have the reader return that instead, unless there is a chance of using the layer interface heights for something else.

Also for WRF-Chem reader it looks like `getvar` will give you the layer interface heights if you request "zstag". So maybe here we should use that.

blychs:
@zmoon I am not sure about the calculations. Do you think that having the layer height could be useful to interpolate to other (intermediate) altitudes? I'm thinking mostly about the aircraft team - I know close to nothing about aircraft field campaigns. If it isn't, then changing this to layer thickness wouldn't be to hard.

mlirenzhenmayi:
@zmoon @blychs PH and PHB are used to calculate the height for each layer, and then we can calculate the partial column for each layer. Pressure for each layer = PB + P, from WRF-Chem. It's used for averaging kernels revision, based on which we established a function of Pressure and averaging kernel for each layer from TROPOMI, interpolate the model pressure to TROPOMI pressure for each layer, and then get the layered averaging kernel for the model.

zmoon:
Hi @mlirenzhenmayi thanks for the comments. Is it correct that PB + P is the pressure of the layer interfaces? To use for conservative vertical interp based on pressure, for example.

@rschwant and I think these calculations of interface height/pressure and layer height/pressure thickness might be better done inside the reader and assigned standardized variable names, like currently done for the mid-layer pressure/height. But that's outside the scope of this PR.

mlirenzhenmayi:
> Hi @mlirenzhenmayi thanks for the comments. Is it correct that PB + P is the pressure of the layer interfaces? To use for conservative vertical interp based on pressure, for example.
> 
> @rschwant and I think these calculations of interface height/pressure and layer height/pressure thickness might be better done inside the reader and assigned standardized variable names, like currently done for the mid-layer pressure/height. But that's outside the scope of this PR.

PB+P is the mid-layer pressure. Maybe it's the tricky part of WRF-Chem. WRF-Chem doesn't provide the standard mid-layer pressure, so we need to calculate it based on the PB and P. If there's a standard calculation for mid-layer pressure covering different models, we can use that instead of calculating separately.

rschwant:
@mlirenzhenmayi I was more thinking that we can move your calculation for these variables into the WRF-Chem reader, since they are model dependent. And then each model will do these calculations differently. And the analysis code will just use the standard value across all the models. Does that make sense? This can be a version 2 satellite advancement once we start standardizing the code more too.

mlirenzhenmayi:
> @mlirenzhenmayi I was more thinking that we can move your calculation for these variables into the WRF-Chem reader, since they are model dependent. And then each model will do these calculations differently. And the analysis code will just use the standard value across all the models. Does that make sense? This can be a version 2 satellite advancement once we start standardizing the code more too.

I think that makes sense. I do have a separate calculation module for this: https://github.com/NOAA-CSL/MELODIES-MONET/blob/develop/melodies_monet/util/cal_mod_no2col.py . It can be straight forward to combine cal_mod_no2col.py with the reader.

zmoon:
@mlirenzhenmayi I think the partial column calculation can be outside the reader (or inside, open to others' thoughts about that).

However, I do think it makes sense for the reader to provide the variables you need, instead of you needing to compute them from WRF-specific P,PB,PH,PHB,T.
* temperature is already provided[^1] as `temperature_k` (computed via T by wrf-python `getvar`)
* pressure is already provided as `pres_pa_mid` (computed via P+PB by wrf-python `getvar`)
* layer $\Delta z$ is not yet provided, we can add this (taking the diff of wrf-python `getvar` `'zstag'`)

[^1]: [When](https://github.com/noaa-oar-arl/monetio/blob/2d4dc68cedde785d44a2623b92c5f127af8c8038/monetio/models/_wrfchem_mm.py#L81) `surf_only_nc` is false.

mlirenzhenmayi:
> @mlirenzhenmayi I think the partial column calculation can be outside the reader (or inside, open to others' thoughts about that).
> 
> However, I do think it makes sense for the reader to provide the variables you need, instead of you needing to compute them from WRF-specific P,PB,PH,PHB,T.
> 
> * temperature is already provided[1](#user-content-fn-1-01fc4329e5191ca1b11f81ba3f1d85d5) as `temperature_k` (computed via T by wrf-python `getvar`)
> * pressure is already provided as `pres_pa_mid` (computed via P+PB by wrf-python `getvar`)
> * layer 
>     Δ
>     z
>    is not yet provided, we can add this (taking the diff of wrf-python `getvar` `'zstag'`)
> 
> ## Footnotes
> 1. [When](https://github.com/noaa-oar-arl/monetio/blob/2d4dc68cedde785d44a2623b92c5f127af8c8038/monetio/models/_wrfchem_mm.py#L81) `surf_only_nc` is false. [↩](#user-content-fnref-1-01fc4329e5191ca1b11f81ba3f1d85d5)

@zmoon The partial column calculation has already been calculated outside of the reader. It's done in cal_mod_no2col.py:  https://github.com/NOAA-CSL/MELODIES-MONET/blob/develop/melodies_monet/util/cal_mod_no2col.py . I agree that the P, PB, PHB and PH should be read by default in the reader.

monetio/sat/_tropomi_l2_no2_mm.py
```diff
@@ -54,45 +60,120 @@ def _open_one_dataset(fname, variable_dict):
...
+        if dct_ is None:
+            dct_ = variable_dict.get(varname_, {})
+        group_name = dct_.get("group", default_group)
+        if isinstance(group_name, list):
+            group_name = group_name[0]
+        return _get_values(dso[group_name][varname_], dct_)
+
+    for varname, dct in variable_dict.items():
+        print(f"- {varname}")
+        if dct is None:
```
mlirenzhenmayi: # from mli: the "None" is read as a str from yaml file, so cannot recognized
if (dct is None) or (dct == "None") or (dct == "NONE"):
            dct = {}

monetio/sat/_tropomi_l2_no2_mm.py
```diff
@@ -103,16 +184,14 @@ def apply_quality_flag(ds):
...
-                values = ds[varname].values
-                values[quality_flag <= quality_thresh_min] = np.nan
+        # Apply the quality thresh minimum to selected variables
+        for varname in ds.attrs["var_applied"]:
+            logging.debug(f"applying quality flag to {varname}")
+            values = ds[varname].values
+            values[quality_flag <= quality_thresh_min] = np.nan
 
 
 def open_dataset(fnames, variable_dict, debug=False):
```
mlirenzhenmayi, line 198: can we change the function name to **read_trpdataset**? Just to be consistent with driver which has been incorporated into melodies-monet

mlirenzhenmayi (changes requested):
I tested the changes with my previous codes, and the results are right. I only have two comments for code revision.

monetio/sat/_tropomi_l2_no2_mm.py
```diff
@@ -54,45 +60,120 @@ def _open_one_dataset(fname, variable_dict):
...
+                    "surface_pressure",
+                    default_group="PRODUCT/SUPPORT_DATA/INPUT_DATA",
+                    dct_=dct,
+                ),
+                dims=("y", "x"),
+            )
+
+            # Mid-layer pressure
+            assert a.sizes["v"] == 2, "base and top"
+            p = ((a.isel(v=0) + b.isel(v=0) * psfc) + (a.isel(v=1) + b.isel(v=1) * psfc)) / 2
```
mlirenzhenmayi, line 99: can you change the dimension from (z,y,x,z) to (y,x,z)? This dimension is consistent with the follow-up process of TROPOMI.

p = p.transpose('y','x','z')

mlirenzhenmayi (changes requested):
Add one comment on pressure dimension

monetio/sat/_tropomi_l2_no2_mm.py
blychs: According to the pyYaml docs (https://pyyaml.org/wiki/PyYAMLDocumentation), None should be represented either with ~ or with null

monetio/sat/_tropomi_l2_no2_mm.py
blychs: @mlirenzhenmayi Would this order be desirable for _camx_mm as well?

mlirenzhenmayi: > @mlirenzhenmayi Would this order be desirable for _camx_mm as well?

I'm really not sure. Have not looked into _camx_mm before. If you're using the same averaging kernel correction processing like TROPOMI-WRF_Chem, then I suppose the dimension should be ('y','x','z') to avoid errors

monetio/sat/_tropomi_l2_no2_mm.py
blychs: According to the pyYaml docs (https://pyyaml.org/wiki/PyYAMLDocumentation), None should be represented either with ~ or with null

monetio/sat/_tropomi_l2_no2_mm.py
blychs: I have only worked on it for TEMPO, but if only that and the layer thickness are needed (we already have there the other functions) it seems to me like an easy change that would get it working with the code you wrote for TROPOMI as well.
If you think that should work, I am happy to do the changes in CAMx and get a more consistent codebase

monetio/sat/_tropomi_l2_no2_mm.py
zmoon: Addressed in 8fee18c2d7393132753c2c9bab63881038753efd

monetio/sat/_tropomi_l2_no2_mm.py
zmoon: I believe that for model data for MM at least, (time, z, y, x) is the "standard".

An alternative approach @mlirenzhenmayi would be to use the `.transpose('y','x','z')` near the start of your processing function, to ensure dims always match your expectation. Or, with xarray it should be possible to write things that don't require a certain dim order.

monetio/sat/_tropomi_l2_no2_mm.py
zmoon: I added it as an alias: 626b95c 

In general, we're trying to keep API's for the different model/sat readers somewhat consistent, with the same function names etc.

monetio/sat/_tropomi_l2_no2_mm.py
mlirenzhenmayi: > I believe that for model data for MM at least, (time, z, y, x) is the "standard".
> 
> An alternative approach @mlirenzhenmayi would be to use the `.transpose('y','x','z')` near the start of your processing function, to ensure dims always match your expectation. Or, with xarray it should be able to write things that don't require a certain dim order.

OK. I can revise the processing module to transpose the dimensions. Everything looks good to me now.

mlirenzhenmayi (approved):

monetio/sat/_tropomi_l2_no2_mm.py
zmoon: Ok, thanks a lot!

PR #185, opened 2024-08-19
Add a _camx_mm reader

blychs:
@zmoon This is my _camx_mm reader. All of the surface plots have been tested with melodies monet (I will submit a PR there too once this one is accepted), using the wrfchem example in the docs.
I have not written a test for this, but it should be doable. The problem is that I'm using data which was given to me, and am not able to make it public, for the test case.
We could use the output provided by CAMx itself if that is permissible.

Cheers
Pablo


zmoon:
> the output provided by CAMx itself if that is permissible

Sure, that sounds promising.

Also, I see two other readers included in this PR. Could you create a new branch/PR that only has your CAMx reader?

blychs:
> > the output provided by CAMx itself if that is permissible
> 
> Sure, that sounds promising.
> 
> Also, I see two other readers included in this PR. Could you create a new branch/PR that only has your CAMx reader?
Sure, I am sorry about that ( I thought I had isolated it already, but didn't notice I had done so over NCAR's fork).
I think that this PR can be closed, I'll create a new one with only that.

PR #186, opened 2024-08-23
feature/_camx_mm

blychs:
_camx_mm reader (redone to avoid updating more things that strictly related to this PR).

I still have not built a test. I have checked the output files that are provided by the CAMx developers, but see the following problems to add that as a test case:
a) The files are provided in a tar.gz bundle, which is about 440 MB. As far as I understand, we wanted to have testcases of less than 200 MB. The file we really need is only 32 MB, but we would need to host it somewhere else. I don't know whether that would be acceptable, and I am uncertain about any licensing issues with doing that. If we also want to have the meteorological variables, they are on a separate file, which is only 24 MB for the 2D files, but 300 MB if we require 3D.
b) The files provided by the CAMx are only 2D, and if we need 3D, we would have to get someone to run CAMx. I have not compiled it, and only used output that was provided to me (I don't think that I am allowed to make that public).
c) CAMx output (and input) can be either netCDF or fortran binaries (UAMIV). My code is able to handle both, but we don't have examples with both available.

Let me know your thoughts.
Pablo


blychs:
I am sorry for this, I don't know why my code is still including things it should not

PR #188, opened 2024-08-23
TEMPO NO2 reader

blychs:
This is a tempo reader based on the TROPOMI NO2 reader, set up for compatibility with MELODIES-MONET. It has been tested with locally downloaded data, using MONET plotting capabilities. Pairing (at the MONET or MELODIES-MONET levels) has not been programmed in yet. A test case for this, with automatic downloads, would require an earthdata user --which would also require an additional dependency (namely, earthdata). If that is acceptable, I could set it up.
Pablo

### DIFF
--- a/monetio/sat/__init__.py
+++ b/monetio/sat/__init__.py
@@ -14,12 +14,12 @@ from . import (
 )
 
 __all__ = [
-    "_tempo_l2_no2_mm",
     "_gridded_eos_mm",
     "_modis_l2_mm",
     "_mopitt_l3_mm",
     "_omps_l3_mm",
     "_omps_nadir_mm",
+    "_tempo_l2_no2_mm",
     "_tropomi_l2_no2_mm",
     "goes",
     "modis_ornl",
--- a/monetio/sat/__init__.py
+++ b/monetio/sat/__init__.py
@@ -4,6 +4,7 @@ from . import (
     _mopitt_l3_mm,
     _omps_l3_mm,
     _omps_nadir_mm,
+    _tempo_l2_no2_mm,
     _tropomi_l2_no2_mm,
     goes,
     modis_ornl,
@@ -13,6 +14,7 @@ from . import (
 )
 
 __all__ = [
+    "_tempo_l2_no2_mm",
     "_gridded_eos_mm",
     "_modis_l2_mm",
     "_mopitt_l3_mm",
--- a/monetio/sat/_tempo_l2_no2_mm.py
+++ b/monetio/sat/_tempo_l2_no2_mm.py
@@ -1,0 +1,146 @@
+"""TEMPO L2 NO2 data file reader.
+
+History:
+
+
+"""
+
+import logging
+import os
+import sys
+from collections import OrderedDict
+from glob import glob
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+from netCDF4 import Dataset
+
+
+def _open_one_dataset(fname, variable_dict):
+    """Read locally stored INTELSAT TEMPO NO2 level 2
+    Parameters
+    ----------
+    fname: string
+        fname is local path to netcdf4 (HDF5) file
+
+    variable_dict: dict
+    Returns
+    -------
+    ds: xarray dataset
+    """
+    print("reading " + fname)
+
+    ds = xr.Dataset()
+
+    dso = Dataset(fname, "r")
+    lon_var = dso.groups["geolocation"]["longitude"]
+    lat_var = dso.groups["geolocation"]["latitude"]
+    time_var = dso.groups["geolocation"]["time"]
+
+    ds["lon"] = (
+        ("x", "y"),
+        lon_var[:].squeeze(),
+        {"long_name": lon_var.long_name, "units": lon_var.units},
+    )
+    ds["lat"] = (
+        ("x", "y"),
+        lat_var[:].squeeze(),
+        {"long_name": lat_var.long_name, "units": lat_var.units},
+    )
+    ds["time"] = (
+        ("time",),
+        time_var[:].squeeze(),
+        {"long_name": time_var.long_name, "units": time_var.units},
+    )
+    ds["lon"] = ds["lon"].fillna(-999.99)
+    ds["lat"] = ds["lat"].fillna(-999.99)
+    ds = ds.set_coords(["time", "lon", "lat"])
+
+    ds.attrs["reference_time_string"] = dso.time_coverage_start
+    ds.attrs["granule_number"] = dso.granule_num
+    ds.attrs["scan_num"] = dso.scan_num
+
+    for varname in variable_dict:
+        values_var = dso.groups["product"][varname]
+        values = values_var[:].squeeze()
+        fv = values_var.getncattr("_FillValue")
+        if not np.isfinite(fv):
+            values[:][values[:] == fv] = np.nan
+
+        if "scale" in variable_dict[varname]:
+            values[:] = variable_dict[varname]["scale"] * values[:]
+
+        if "minimum" in variable_dict[varname]:
+            minimum = variable_dict[varname]["minimum"]
+            values[:][values[:] < minimum] = np.nan
+
+        if "maximum" in variable_dict[varname]:
+            maximum = variable_dict[varname]["maximum"]
+            values[:][values[:] > maximum] = np.nan
+
+        ds[varname] = (("x", "y"), values, values_var.__dict__)
+
+        if "quality_flag_max" in variable_dict[varname]:
+            ds.attrs["quality_flag"] = varname
+            ds.attrs["quality_thresh_max"] = variable_dict[varname]["quality_flag_max"]
+
+    dso.close()
+
+    return ds
+
+    # time_var = dso.groups["geolocation"]["time"]
+
+
+def apply_quality_flag(ds):
+    """Mask variables in place based on quality flag
+    Parameters
+    ----------
+    ds: xr.Dataset
+    """
+    if "quality_flag" in ds.attrs:
+        quality_flag = ds[ds.attrs["quality_flag"]]
+        quality_thresh_max = ds.attrs["quality_thresh_max"]
+
+        # Apply the quality thersh maximum to all variables in ds
+        for varname in ds:
+            if varname != ds.attrs["quality_flag"]:
+                logging.debug(varname)
+                values = ds[varname].values
+                values[quality_flag > quality_thresh_max] = np.nan
+
+
+def open_dataset(fnames, variable_dict, debug=False):
+    if debug:
+        logging_level = logging.DEBUG
+        logging.basicConfig(strea=sys.stdout, level=logging_level)
+
+    if isinstance(fnames, Path):
+        fnames = fnames.as_posix()
+    if isinstance(variable_dict, str):
+        variable_dict = {variable_dict: {}}
+    elif isinstance(variable_dict, dict):
+        pass
+    else:  # Assume sequence
+        variable_dict = {varname: {} for varname in variable_dict}
+
+    for subpath in fnames.split("/"):
+        if "$" in subpath:
+            envvar = subpath.replace("$", "")
+            envval = os.getenv(envvar)
+            if envval is None:
+                raise Exception("Environment variable not defined: " + subpath)
+            else:
+                fnames = fnames.replace(subpath, envval)
+
+    print(fnames)
+    files = sorted(glob(fnames))
+
+    granules = OrderedDict()
+
+    for file in files:
+        granule = _open_one_dataset(file, variable_dict)
+        apply_quality_flag(granule)
+        granules[granule.attrs["reference_time_string"]] = granule
+
+    return granules
--- a/monetio/sat/_tempo_l2_no2_mm.py
+++ b/monetio/sat/_tempo_l2_no2_mm.py
@@ -8,10 +8,10 @@ History:
 import logging
 import os
 import sys
+import warnings
 from collections import OrderedDict
 from glob import glob
 from pathlib import Path
-import warnings
 
 import numpy as np
 import xarray as xr
@@ -63,8 +63,10 @@ def _open_one_dataset(fname, variable_dict):
     ds.attrs["scan_num"] = dso.scan_num
 
     if ("pressure" in variable_dict) and "surface_pressure" not in variable_dict:
-        warnings.warn("Calculating pressure in TEMPO data requires surface_pressure. "
-                     + "Adding surface_pressure to output variables")
+        warnings.warn(
+            "Calculating pressure in TEMPO data requires surface_pressure. "
+            + "Adding surface_pressure to output variables"
+        )
         variable_dict["surface_pressure"] = {}
 
     for varname in variable_dict:
@@ -171,7 +173,7 @@ def calculate_pressure(ds):
     n_layers = len(surf_pressure.Eta_A)
     press = np.zeros((surf_pressure.shape[0], surf_pressure.shape[1], n_layers))
     for k in range(0, n_layers):
-        press[:, :, k] = (eta_a[k] + eta_b[k] * surf_pressure.values)
+        press[:, :, k] = eta_a[k] + eta_b[k] * surf_pressure.values
     pressure = xr.DataArray(
         data=press,
         dims=("x", "y", "swt_level_stagg"),
--- a/monetio/sat/_tempo_l2_no2_mm.py
+++ b/monetio/sat/_tempo_l2_no2_mm.py
@@ -12,10 +12,10 @@ import warnings
 from collections import OrderedDict
 from glob import glob
 from pathlib import Path
-from cftime import num2pydate
 
 import numpy as np
 import xarray as xr
+from cftime import num2pydate
 from netCDF4 import Dataset
 
 
@@ -40,7 +40,6 @@ def _open_one_dataset(fname, variable_dict):
     lat_var = dso.groups["geolocation"]["latitude"]
     time_var = dso.groups["geolocation"]["time"]
     time_units = time_var.units
-    #import pdb; pdb.set_trace()
 
     ds["lon"] = (
         ("x", "y"),
@@ -57,7 +56,6 @@ def _open_one_dataset(fname, variable_dict):
         num2pydate(time_var[:].squeeze(), time_units),
         {"long_name": time_var.long_name, "units": time_var.units},
     )
-    #ds["time"] = num2pydate(ds["time"], ds["time"].units)
     ds = ds.set_coords(["time", "lon", "lat"])
 
     ds.attrs["reference_time_string"] = dso.time_coverage_start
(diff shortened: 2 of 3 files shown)


zmoon:
@blychs , alternatively, if you have an example file, we could place it here: https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_observation_data/satellite/

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,146 @@
...
+            if varname != ds.attrs["quality_flag"]:
+                logging.debug(varname)
+                values = ds[varname].values
+                values[quality_flag > quality_thresh_max] = np.nan
+
+
+def open_dataset(fnames, variable_dict, debug=False):
+    if debug:
+        logging_level = logging.DEBUG
+        logging.basicConfig(strea=sys.stdout, level=logging_level)
```
zmoon:
```suggestion
        logging.basicConfig(stream=sys.stdout, level=logging_level)
```

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,146 @@
...
+
+    ds.attrs["reference_time_string"] = dso.time_coverage_start
+    ds.attrs["granule_number"] = dso.granule_num
+    ds.attrs["scan_num"] = dso.scan_num
+
+    for varname in variable_dict:
+        values_var = dso.groups["product"][varname]
+        values = values_var[:].squeeze()
+        fv = values_var.getncattr("_FillValue")
+        if not np.isfinite(fv):
```
zmoon: What is the rationale behind this condition? `_FillValue` could be set to anything.

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,146 @@
...
+        lat_var[:].squeeze(),
+        {"long_name": lat_var.long_name, "units": lat_var.units},
+    )
+    ds["time"] = (
+        ("time",),
+        time_var[:].squeeze(),
+        {"long_name": time_var.long_name, "units": time_var.units},
+    )
+    ds["lon"] = ds["lon"].fillna(-999.99)
+    ds["lat"] = ds["lat"].fillna(-999.99)
```
zmoon: Why not leave them as NaN?

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,146 @@
...
+    return ds
+
+    # time_var = dso.groups["geolocation"]["time"]
+
+
+def apply_quality_flag(ds):
+    """Mask variables in place based on quality flag
+    Parameters
+    ----------
+    ds: xr.Dataset
```
zmoon:
```suggestion

    Parameters
    ----------
    ds : xr.Dataset
```

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,146 @@
...
+    """Read locally stored INTELSAT TEMPO NO2 level 2
+    Parameters
+    ----------
+    fname: string
+        fname is local path to netcdf4 (HDF5) file
+
+    variable_dict: dict
+    Returns
+    -------
+    ds: xarray dataset
```
zmoon:
```suggestion
    Parameters
    ----------
    fname : str
        Local path to netCDF4 (HDF5) file.
    variable_dict : dict

    Returns
    -------
    ds : xr.Dataset
```
more info: https://numpydoc.readthedocs.io/en/latest/format.html

monetio/sat/_tempo_l2_no2_mm.py
blychs: Because matplotlib.pcolormesh breaks if there are NaN values in the coordinates (the _values_ can be NaN, but not the coordinates). However, the lat/lon that are NaN in the original arrays are because the satellite is pointing at space, and therefore lat and lon are not defined.


 Another solution might be using pcolor, which is the oficially suggested solution (https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pcolormesh.html, look at the section on masked arrays), but that would break away from the behavior of draw_map() in MONET. 
The workaround of turning them into -999.99 was taken from Amy K. Huff (https://gist.github.com/akhuff157/8411788ed06d459972f9b15203c8d793), but if you have a better solution, I am happy to change this.

EDIT: Clarify the pcolor suggested solution
EDIT2: correct typo

blychs:
> @blychs , alternatively, if you have an example file, we could place it here: https://csl.noaa.gov/groups/csl4/modeldata/melodies-monet/data/example_observation_data/satellite/

I do have files for this. Let me know how to give them to you so that they can be uploaded (I would just need to get rid of unnecessary variables so that we can have a full scan).

monetio/sat/_tempo_l2_no2_mm.py
zmoon: Since this issue is more to do with plotting, I would prefer this not be done in the reader. Perhaps it could be done in MM before pcolormesh is used. Or, we could do it here but make it an option that is off by default.

zmoon:
> I do have files for this. Let me know how to give them to you so that they can be uploaded (I would just need to get rid of unnecessary variables so that we can have a full scan).

Sounds good. I'll send an email.

blychs:
I found a bug in my code, which is with the application of the quality flags, maximum and minimum. i. e., I'm asking it to compare the values with the max, min or (for the quality_flag) qa_treshold. However, this breaks if the value is NaN, due to the way the comparison is made. It should be an easy fix, which I'll hopefully update today.

blychs:
Those bugs should be fixed now. I also added options for the rest of the possible variables a user might want, including everything needed to apply air mass factors and scattering weights and a pressure calculations. This means that the changes are quite large, and might require careful checking. I tried my code with all variables and it seems to be working fine, and the pre-commit hooks are passing. I don't really know how to run the linter though, should I just install pylint and give it a go?

zmoon:
>  the pre-commit hooks are passing. I don't really know how to run the linter though

Linting is included in the pre-commit hooks. But if you do want to run it (flake8) individually sometime:
```
pre-commit run flake8 --all-files
```
or
```
pre-commit run flake8 --file some/path.py
```

monetio/sat/_tempo_l2_no2_mm.py
zmoon: For example, the default nc fill value for double, even though it's large, is still finite, so the replacement by NaN will be skipped with this current `if not np.isfinite(fv):` guard.
```python
>>> np.isfinite(netCDF4.default_fillvals["f8"])
True
```

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,233 @@
...
+        fv = values_var.getncattr("_FillValue")
+        if not np.isfinite(fv):
+            values[:][values[:] == fv] = np.nan
+
+        if "scale" in variable_dict[varname]:
+            values[:] = variable_dict[varname]["scale"] * values[:]
+
+        if "minimum" in variable_dict[varname]:
+            minimum = variable_dict[varname]["minimum"]
+            values[:] = np.where(values[:] >= minimum, values[:], np.nan)
```
zmoon: With this change, if `minimum` is NaN, `values` gets fully filled with NaN (whereas before it would do nothing). Is that really the desired behavior? Why would one set minimum/maximum to NaN anyway?

blychs: The problem before was when minimum was a number (either int or float) and some values were NaN: the code would just break saying that it could not do the comparison, at least with Python 3.11. 

I hadn't realized the problem of minimum being NaN (no one should set it to NaN, but I guess it still needs to be addressed), and was addressing the case of:
a) minimum not being there (in which case the np.where() function is not applied),
b) minimum being a number, in which case this works.

I'll add a check for minimum being NaN to avoid this.

zmoon: > The problem before was when minimum was a number (either int or float) and some values were NaN: the code would just break saying that it could not do the comparison, at least with Python 3.11.

I tried to replicate this with Python 3.11 / NumPy 2.1.0 and was not able to. Is this not the correct situation?
```python
>>> a = np.r_[1.0, 2.0, np.nan, 4.0]
>>> a[a < 2] = np.nan
>>> a
array([nan,  2., nan,  4.])
```

blychs: OK, this is weird. I only get the bug I mentioned with the "terrain height" variable, and I don't know why.
Here is the full Traceback in an Ipython session.
It might be related to the datatype of terrain height. It is likely that that variable is not really needed, but I would rather have it as an option (since a user might want to know the height of the terrain at a certain location according to TEMPO).

```
Cell In[2], line 1
----> 1 vals = mio.sat._tempo_l2_no2_mm.open_dataset(
      2     "/glade/campaign/acom/acom-da/sma/TEMPO_NO2_V03/TEMPO_NO2_L2_V03_20240707*S011G02*.nc",
      3     {
      4         "vertical_column_stratosphere": {"minimum":0},
      5         "scattering_weights": {},
      6         "vertical_column_troposphere": {},
      7         "main_data_quality_flag": {"quality_flag_max": 0},
      8         #"surface_pressure": {"maximum": 100000},
      9         "terrain_height": {"minimum": 0}
     10     },
     11 )

File ~/melodies-monet/monetio/monetio/sat/_tempo_l2_no2_mm.py:230, in open_dataset(fnames, variable_dict, debug)
    227 granules = OrderedDict()
    229 for file in files:
--> 230     granule = _open_one_dataset(file, variable_dict)
    231     apply_quality_flag(granule)
    232     granules[granule.attrs["reference_time_string"]] = granule

File ~/melodies-monet/monetio/monetio/sat/_tempo_l2_no2_mm.py:118, in _open_one_dataset(fname, variable_dict)
    116     minimum = variable_dict[varname]["minimum"]
    117     #values[:] = np.where(values[:] >= minimum, values[:], np.nan)
--> 118     values[values < minimum] = np.nan
    120 if "maximum" in variable_dict[varname]:
    121     maximum = variable_dict[varname]["maximum"]

File /glade/work/plichtig/conda-envs/melodies-monet-test/lib/python3.11/contextlib.py:81, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     78 @wraps(func)
     79 def inner(*args, **kwds):
     80     with self._recreate_cm():
---> 81         return func(*args, **kwds)

File /glade/work/plichtig/conda-envs/melodies-monet-test/lib/python3.11/site-packages/numpy/ma/core.py:3387, in MaskedArray.__setitem__(self, indx, value)
   3384     mval = tuple([False] * len(_dtype.names))
   3385 if _mask is nomask:
   3386     # Set the data, then the mask
-> 3387     _data[indx] = dval
   3388     if mval is not nomask:
   3389         _mask = self._mask = make_mask_none(self.shape, _dtype)

ValueError: cannot convert float NaN to integer
```

EDIT: I con confirm that this is the case:

```
In [4]: np.nan < 3
Out[4]: False

In [5]: a = np.array([1,2,3,4,5], dtype=int)

In [6]: 

In [6]: a<np.nan
Out[6]: array([False, False, False, False, False])

In [7]: a[a<np.nan] = np.nan
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[7], line 1
----> 1 a[a<np.nan] = np.nan

ValueError: cannot convert float NaN to integer
```

This does not happen for floats.
I'll work out a way to deal with this and update this again

zmoon: Weird that terrain height would be int. But yes I see the issue. Easiest solution would be to convert the array to float first (`.astype(float)`) before NaN application. But it looks this may actually be a masked array, so you'd really want to modify the mask, not set values to NaN.

Thanks for working on this, which relates to #184 as well. I encountered a masked int variable [there](https://github.com/noaa-oar-arl/monetio/blob/0a93c4396057257d654d75e44b08ba49d34a1618/monetio/sat/_tropomi_l2_no2_mm.py#L151) too and was just skipping it, but maybe `.astype(float).filled(np.nan)` conversion should be done instead.

blychs: Casting it is definitely a solution, but another (possibly simpler) solution is just inverting the condition of my previous code. You can look at my next commit for that. I. e., it is doing


values = np.where(values [:] < min, np.nan, values[:])

Since the condition is always false if min==np.nan , this would avoid the previous issue, and wouldn't require casting the array into float.

Although, of course, this is still modifying the values instead of the mask. I should probably take your advice and look at how to modify the mask.
terrain_height is of type "short" (ie, float32), so I'm not sure why it fails as if it were int, but it is the exact same error message. Maybe it's a "feature" of masked arrays?

blychs: Done. Now it deals with "maximum" and "minimum" by masking the arrays. I don't know if this really makes a difference at the xarray.DataArray level.

monetio/sat/_tempo_l2_no2_mm.py
blychs: Let's do it at the MM level.
I removed those lines in [fb4f571](https://github.com/noaa-oar-arl/monetio/pull/188/commits/fb4f571e86dbd3348b61eeb6904435a335c6d5d0)

monetio/sat/_tempo_l2_no2_mm.py
blychs: Done.

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,246 @@
...
+            "amf_troposphere",
+            "amf_stratosphere",
+        ]:
+            values_var = dso.groups["support_data"][varname]
+        elif varname in ["fit_rms_residual", "fit_convergence_flag"]:
+            values_var = dso.groups["qa_statistics"][varname]
+        values = values_var[:].squeeze()
+        fv = values_var.getncattr("_FillValue")
+        if not np.isfinite(fv):
+            values[:][values[:] == fv] = np.nan
```
zmoon:
```suggestion
```
From what I've read, this is done automatically by netcdf4-python (though it uses masked array to do it), so no need to redo it here.

tests/test_tempo_l2.py
```diff
@@ -0,0 +1,80 @@
...
+    assert set(ds2.variables) == {
+        "lat",
+        "lon",
+        "main_data_quality_flag",
+        "pressure",
+        "surface_pressure",
+        "time",
+        "vertical_column_troposphere",
+    }
+    assert set(ds2["pressure"].dims) == {"swt_level_stagg", "x", "y"}
```
zmoon:
```suggestion
    assert ds2["pressure"].dims == ("swt_level_stagg", "x", "y")
```
tuple to test order too

tests/test_tempo_l2.py
```diff
@@ -0,0 +1,80 @@
...
+def test_open_dataset(test_file_path):
+    vn = "vertical_column_troposphere"
+    t_ref = "2024-08-26T20:40:05Z"
+    ds = open_dataset(test_file_path, {vn: {}})[t_ref]
+
+    assert set(ds.coords) == {"time", "lat", "lon"}
+    assert set(ds) == {vn}
+    assert set(ds.attrs) == {"granule_number", "reference_time_string", "scan_num"}
+
+    with pytest.warns(UserWarning):
```
zmoon: Can you add a `match="..."` to be more specific about the warning being captured?

tests/test_tempo_l2.py
blychs: Currently they are not in that order, but I'm happy to change that to make it consistent with other readers

monetio/sat/__init__.py
```diff
@@ -13,6 +14,7 @@
 )
 
 __all__ = [
+    "_tempo_l2_no2_mm",
```
zmoon: Could you move this to the alpha order position like in the imports?

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+    ds["lat"] = (
+        ("x", "y"),
+        lat_var[:].squeeze(),
+        {"long_name": lat_var.long_name, "units": lat_var.units},
+    )
+    ds["time"] = (
+        ("time",),
+        num2pydate(time_var[:].squeeze(), time_units),
+        {"long_name": time_var.long_name, "units": time_var.units},
```
zmoon:
```suggestion
        {"long_name": time_var.long_name},
```
Units attr no longer relevant after decoding

blychs: I think that the units attribute is worth keeping. However, the way it's now fails at save when it tries to encode it, because it is creating a DatetimeIndex with it's own encoding.
I'll find a way to fix it and commit again. Sorry for this.

zmoon: You could call it "original_units" instead?

tests/test_tempo_l2.py
zmoon: You could change the test instead, either way

monetio/sat/_tempo_l2_no2_mm.py
blychs: The units need to be the real units, that will be found in the attributes of the netcdf that gets created if someone saves it. The problems seems to lie in the fact that xarray is creating and index with time, not just the (expected) coordinates and variable. I need to find out a way to get rid of that.
Normally, with xarray, if you set valid units for time, it should be able to take care of applying the correct encoding on its own.

blychs: OK, you were right and I was wrong about this. XArray manages the encoding on it's own, and even though I can force it (by setting an ds["time"].encoding = {"dtype":  "float32", "units": time_var.units} ) it doesn't seem to make a lot of sense. The only *nicer* thing about doing that is for people to might save de file and want to process it with other tools, that might need consistency in the time units.
Just going with your commit is the right way to go, and I just leave this here in case we need to refer to it in the future.

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,246 @@
...
+
+import numpy as np
+import xarray as xr
+from cftime import num2pydate
+from netCDF4 import Dataset
+
+
+def _open_one_dataset(fname, variable_dict):
+    """Read locally stored INTELSAT TEMPO NO2 level 2
+    Parameters
```
zmoon, line 25:
```suggestion

    Parameters
```

monetio/sat/_tempo_l2_no2_mm.py
```diff
@@ -0,0 +1,246 @@
...
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+
+    Returns
+    -------
+    layer_pressure: xr.DataArray
+        Pressure at layer in Pa
+    delta_pressure: xr.DataArray
```
zmoon:
```suggestion
    layer_pressure : xr.DataArray
        Pressure at layer in Pa
    delta_pressure : xr.DataArray
```

zmoon (approved):
Thanks @blychs for your work on this and helping with #184 as well.

PR #192, opened 2024-09-11
Fix bugs in layer height and temperature

blychs:
Here are the fixes for the bug I mentioned.
I am sorry for having to reopen a pull request about this!
In short:
the previous code only worked if CAMX had NETCDF input files, which is not always the case, and had a typo at calling for dset_variables. It was also not adding the temperature.

I also changed the name of layer height to make it more clear.

Cheers
Pablo

### DIFF
--- a/monetio/models/_camx_mm.py
+++ b/monetio/models/_camx_mm.py
@@ -440,12 +440,6 @@ def _calc_midlayer_height_agl(dset):
         DataArray with the midlayer height above ground level
     """
 
-    #     assert dset["z"].dims == (
-    #         "TSTEP",
-    #         "LAY",
-    #         "ROW",
-    #         "COL",
-    #     ), "Check dims of z, should be [TSTEP, LAY, ROW, COL]"
 
     if "z" in dset.variables:
         height = "z"
--- a/monetio/models/_camx_mm.py
+++ b/monetio/models/_camx_mm.py
@@ -75,7 +75,9 @@ def open_mfdataset(
             if "temperature_k" in dset.variables:
                 var_list = var_list + ["temperature_k"]
         else:
-            warnings.warn("Filename for meteorological input not provided. Adding only altitude.")
+            warnings.warn(
+                "Filename for meteorological input not provided. Adding only altitude."
+            )
         if (landuse_file is not None) and ("alt_agl_m_mid" in dset.variables):
             file_keywords = _choose_xarray_engine_and_keywords(landuse_file)
             with xr.open_dataset(**file_keywords) as dset_lu:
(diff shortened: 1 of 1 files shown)


monetio/models/_camx_mm.py
```diff
@@ -479,7 +482,7 @@ def _calc_midlayer_height_msl(dset, dset_lu):
     if "alt_agl_m_mid" in dset.keys():
         alt_agl_m_mid = dset["alt_agl_m_mid"]
     else:
-        alt_agl_m_mid, _ = _calc_midlayer_height_agl(dset)
+        alt_agl_m_mid = _calc_midlayer_height_agl(dset)
```
zmoon: regression? (I think we fixed this)

monetio/models/_camx_mm.py
```diff
@@ -448,15 +448,18 @@ def _calc_midlayer_height_agl(dset):
     else:
         raise "No height variable found, but _calc_midlayer_height_agl was called."
     mid_layer_height = np.array(dset[height])  # height in the layer upper interface of each layer
-    layer_height = dset[height]
+    layer_height_agl = dset[height]
+    layer_height_agl.attrs["long_name"] = "layer_height_agl"
```
zmoon: `long_name` shouldn't generally be the same as `name` (if so no need to add it)

monetio/models/_camx_mm.py
```diff
@@ -448,15 +448,18 @@ def _calc_midlayer_height_agl(dset):
...
+    layer_height_agl.attrs["var_desc"] = "Layer height above ground level at top"
```
zmoon, line 452: The calculation suggests to me that this is all the interface heights (nlayers + 1), not just the tops?

blychs: It's not, because in CAMx height is included as a variable per layer, which only has the altitude at the top.
the height of the surface is in a different variable (called topo or TOPO_M), which is even in a different file.
In that sense, it is quite different from what I am used to in other models.

monetio/models/_camx_mm.py
blychs: You are right. Fixed now.

monetio/models/_camx_mm.py
blychs: Hopefully is better now.

zmoon:
The recent commit displeased the formatter

blychs:
@zmoon Just run pre-commit and black found some issues. Should be fixed now.

zmoon (approved):

blychs:
I just saw that the docs failed. I don't think I modified anything in them?

zmoon:
It was in the link check, CRN website currently down https://www.ncei.noaa.gov/access/crn/

zmoon:
https://www.ncei.noaa.gov/ is back now, but data stuff still out (giving 404)

PR #195, opened 2024-09-19
Fix staggered height var name

### LINKED ISSUE #194: wrf_chem_mm reader loading two different things as height
@zmoon @mlirenzhenmayi 
The _wrfchem_mm.py reader seems to be loading a variable called "height" twice (when var is 'height' and when var is 'zstag'). This causes xr.merge() to fail, at least in my system.
I think that the issue is that, in the version I installed (with pip after cloning the main branch on github, so it should be the latest), zstag is also called "height", although it has bottom_top_stag as vertical dimension.
Easy solution: adding a rename for zstag:
```
  101         elif var in {"height", "height_agl", "zstag"}:                                  
  102             var_wrf = getvar(                                                           
  103                 wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="m" 
  104             )
```
to:
```
  101         elif var in {"height", "height_agl", "zstag"}:                                  
  102             var_wrf = getvar(                                                           
  103                 wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="m" 
  104             )                                                                           
  105             if var == "zstag":                                                          
  106                 var_wrf = var_wrf.rename("zstag")  
```

This should work also with older versions.
Cheers
Pablo

zmoon:
Closes #194

### DIFF
--- a/monetio/models/_wrfchem_mm.py
+++ b/monetio/models/_wrfchem_mm.py
@@ -102,6 +102,8 @@ def open_mfdataset(
             var_wrf = getvar(
                 wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False, units="m"
             )
+            if var == "zstag":
+                var_wrf = var_wrf.rename("zstag")
         else:
             var_wrf = getvar(wrflist, var, timeidx=ALL_TIMES, method="cat", squeeze=False)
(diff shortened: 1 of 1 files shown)


blychs (approved):

zmoon:
Thanks @blychs

PR #196, opened 2024-09-20
Add capability of reading list of files

blychs:
This is useful for looping and iterations. It makes it a lot easier to select files by filename in MM and only load those needed for a certain iteration using this reader.


blychs:
Closing as this is really not needed

PR #197, opened 2024-10-02
Hotfix/_camx_mm consistent with _wrfchem_mm

blychs:
This makes the camx_mm reader consistent with wrfchem_mm by incorporating "dz_m" instad of "layer_height_agl"
We had talked at the time of adapting it once the new wrfchem_mm reader was accepted.

Cheers
Pablo

### DIFF
--- a/monetio/models/_camx_mm.py
+++ b/monetio/models/_camx_mm.py
@@ -68,8 +68,8 @@ def open_mfdataset(
                 dset = add_met_data_3D(dset, dset_met)
             if "alt_agl_m_mid" in dset.variables:
                 var_list = var_list + ["alt_agl_m_mid"]
-            if "layer_height_agl" in dset.variables:
-                var_list = var_list + ["layer_height_agl"]
+            if "dz_m" in dset.variables:
+                var_list = var_list + ["dz_m"]
             if "pres_pa_mid" in dset.variables:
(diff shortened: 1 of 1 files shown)


zmoon (commented):
@blychs I'm a bit confused. 

Does your `layer_height_agl` correspond to `alt_agl_m_mid` in the WRF-Chem reader?
`dz_m` is the thickness of each layer, not height above ground level.

https://github.com/noaa-oar-arl/monetio/blob/8c7af1c1a7cfb81cafb9273f0e6eb0996625f259/monetio/models/_wrfchem_mm.py#L213-L215

blychs:
No, it doesn't. alg_agl_m_mid is the exact same thing here as in _wrfchem_mm. layer_height_agl was the layer height at the interface, which I was using later to create the thickness to calculate partial columns.
SInce the wrfchem reader went a different way, by calculating layer thickness in the reader, I am changing it to do the same, so that future code will work for both.
The TEMPO tool can already deal with the way that wrfchem does it.

blychs:
That way, the TROPOMI tool (or any other) should work for both regardless, since the version loaded by MONETIO will look the same.

zmoon (approved):

zmoon:
The NCDC website still having an issue, causing the test failures.

PR #199, opened 2024-10-04
Small changes to OMPS nadir mapper level 2 reader

mbruckner-work:
In testing the fix to the Melodies Monet driver I came across a small change I made to the OMPS reader that allows for using a list of files instead of expanding a string with glob to get the satellite files to read in. This change is necessary for processing over time intervals in Melodies Monet.

### DIFF
--- a/monetio/sat/_omps_nadir_mm.py
+++ b/monetio/sat/_omps_nadir_mm.py
@@ -37,6 +37,8 @@ def read_OMPS_nm(files):
                 else:
                     data_array = xr.concat([data_array, data], "x")
             except (KeyError, ValueError) as e:
+                # KeyError occurs when file exists but contains no data
+                # ValueError occurs when file cross-track dimensions are different than other files loaded
                 print(f"warning: skipping {filename}. {type(e).__name__} occured: {e}")
     if count == 0:
         raise RuntimeError(f"no files loaded from files={files}")
--- a/monetio/sat/_omps_nadir_mm.py
+++ b/monetio/sat/_omps_nadir_mm.py
@@ -38,6 +38,8 @@ def read_OMPS_nm(files):
                     data_array = xr.concat([data_array, data], "x")
             except (KeyError, ValueError) as e:
                 print(f"warning: skipping {filename}. {type(e).__name__} occured: {e}")
+    if count == 0:
+        raise RuntimeError(f"no files loaded from files={files}")
     return data_array
 
 
--- a/monetio/sat/_omps_nadir_mm.py
+++ b/monetio/sat/_omps_nadir_mm.py
@@ -29,18 +29,23 @@ def read_OMPS_nm(files):
         if isinstance(files,str):
             filelist = sorted(glob(files, recursive=False))
         else: 
-            filelist = files # assume list
+            filelist = sorted(files) # assume list
         for filename in filelist:
             try:
                 data = extract_OMPS_nm(filename)
-
+                #print(data.dims)
                 if count == 0:
                     data_array = data
                     count += 1
                 else:
                     data_array = xr.concat([data_array, data], "x")
-            except KeyError:
+            except KeyError as e:
                 print(f'KeyError occured for: {filename}')
+                print(e)
+            except ValueError as e:
+                print(f'ValueError occured for: {filename}')
+                print(e)
+            #print(data_array.dims)
     return data_array
 
 
@@ -158,4 +163,5 @@ def extract_OMPS_nm(fname):
         },
         attrs={"missing_value": -999},
     )
+    print(ds['time'][0],ds.dims)
     return ds
--- a/monetio/sat/_omps_nadir_mm.py
+++ b/monetio/sat/_omps_nadir_mm.py
@@ -37,8 +37,8 @@ def read_OMPS_nm(files):
                 else:
                     data_array = xr.concat([data_array, data], "x")
             except (KeyError, ValueError) as e:
-                # KeyError occurs when file exists but contains no data
-                # ValueError occurs when file cross-track dimensions are different than other files loaded
+                # KeyError occurs in load when file exists but contains no data
+                # ValueError occurs in concat when file cross-track dimensions are different than other files loaded
                 print(f"warning: skipping {filename}. {type(e).__name__} occured: {e}")
     if count == 0:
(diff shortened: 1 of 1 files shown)


monetio/sat/_omps_nadir_mm.py
```diff
@@ -13,30 +13,36 @@ def read_OMPS_nm(files):
...
-            else:
-                data_array = xr.concat([data_array, data], "x")
-
+            try:
+                data = extract_OMPS_nm(filename)
+                if count == 0:
+                    data_array = data
+                    count += 1
+                else:
+                    data_array = xr.concat([data_array, data], "x")
```
zmoon, line 38: Can you move this block to an `else:` branch below the `except ...:`? This way `count` is only incremented if the try succeeds.

monetio/sat/_omps_nadir_mm.py
```diff
@@ -13,30 +13,36 @@ def read_OMPS_nm(files):
...
+            except KeyError as e:
+                print(f'KeyError occured for: {filename}')
+                print(e)
+            except ValueError as e:
+                print(f'ValueError occured for: {filename}')
+                print(e)
     return data_array
```
zmoon, line 45:
```suggestion
    if count == 0:
        raise RuntimeError(f"no files loaded from files={files}")
    return data_array
```
Without something like this, `data_array` would be unbound if all the tries gave `KeyError` or `ValueError` and were skipped (or if `filelist` was empty), which would give a more confusing error message.

monetio/sat/_omps_nadir_mm.py
```diff
@@ -13,30 +13,36 @@ def read_OMPS_nm(files):
     import xarray as xr
 
     count = 0
-    print(files)
+    #print(files)
```
zmoon:
```suggestion
```

zmoon (commented):
Curious why you are expecting some files to fail to load. Does `extract_OMPS_nm` need modifications?

monetio/sat/_omps_nadir_mm.py
```diff
@@ -13,30 +13,35 @@ def read_OMPS_nm(files):
...
+                    data_array = data
+                    count += 1
+                else:
+                    data_array = xr.concat([data_array, data], "x")
+            except KeyError as e:
+                print(f'KeyError occured for: {filename}')
+                print(e)
+            except ValueError as e:
+                print(f'ValueError occured for: {filename}')
+                print(e)
```
zmoon:
```suggestion
            except (KeyError, ValueError) as e:
                print(f"warning: skipping {filename}. {type(e).__name__} occured: {e}")
```
I thought I had this in my review but it seems to have disappeared

mbruckner-work: Yeah, I went to test a clean install and got error messages for reading in my downloaded test file set that I would have already dealt with.

The KeyError exception exists because I've run into files that contain no data and are essentially just a filename. The output message is "Unable to synchronously open object (object 'GeolocationData' doesn't exist)".   
I also came across an OMPS observation file where the y-dimension (satellite cross track) was very wrong for some reason and should be skipped.

monetio/sat/_omps_nadir_mm.py
mbruckner-work: Thanks for the suggestion. This is likely a more informative place to let the user know there were no files loaded.

monetio/sat/_omps_nadir_mm.py
mbruckner-work: `count` functions as a flag for initializing the combined dataset for all observations to be read in. Unless I'm mistaken, in its current location it only increments for the first file successfully read in.

zmoon: Ah yes I see what you're saying. I would still move this stuff since it's really the `data = extract_OMPS_nm(filename)` that you're `try`ing and want to handle exceptions for. A better option though might be to have a list of data arrays that you append to upon successful load, then just call xarray concat once at the end. Then `count` isn't needed.

zmoon: But fine to leave it this way too.

mbruckner-work: @zmoon: I think that this may need to stay how it is until after the tutorial. The ValueError that comes up for files with weird cross-track dimensions occurs in the xarray concat, so appending on successful load and then calling xarray concat at the very end results in passing no data out of the reader if there is a ValueError.

zmoon: Would you mind adding some comments about the `KeyError` and `ValueError` situations under the `except` line? Just short versions of your comment here about the `ValueError` and the one [here](https://github.com/noaa-oar-arl/monetio/pull/199#discussion_r1788230853).

mbruckner-work: Added.

https://github.com/mbruckner-work/monetio/blob/8c027520e7ec61f111aa524594abc129a0cd4401/monetio/sat/_omps_nadir_mm.py#L39-L42

zmoon: Great, thanks

monetio/sat/_omps_nadir_mm.py
```diff
@@ -13,30 +13,35 @@ def read_OMPS_nm(files):
...
+            try:
+                data = extract_OMPS_nm(filename)
+                if count == 0:
+                    data_array = data
+                    count += 1
+                else:
+                    data_array = xr.concat([data_array, data], "x")
+            except (KeyError, ValueError) as e:
+                # KeyError occurs when file exists but contains no data
+                # ValueError occurs when file cross-track dimensions are different than other files loaded
```
zmoon: Correct?
```suggestion
                # KeyError occurs in load when file exists but contains no data
                # ValueError occurs in concat when file cross-track dimensions are different than other files loaded
```

mbruckner-work: Yes.

zmoon:
@mbruckner-work I can go ahead and resolve the linter/formatter stuff if you want.

mbruckner-work:
@zmoon That would be great. Thanks!

zmoon (approved):
@mbruckner-work If you're satisfied, I will merge this. It's just the ISH tests failing afaict.

PR #202, opened 2024-10-07
Skip ISH tests if the root pages unreachable

zmoon:
To get CI ✔ again. NCEI servers still having issues.

### DIFF
--- a/tests/test_ish_lite.py
+++ b/tests/test_ish_lite.py
@@ -3,6 +3,14 @@ import pytest
 
 from monetio import ish_lite
 
+try:
+    import requests
+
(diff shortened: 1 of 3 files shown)


tests/test_ish_lite.py
```diff
@@ -3,6 +3,14 @@
 
 from monetio import ish_lite
 
+try:
+    import requests
+
+    r = requests.head("https://www1.ncdc.noaa.gov/pub/data/noaa/isd-lite/")
+    r.raise_for_status()
```
zmoon: Doesn't seem to be needed for HEAD.
```suggestion
```

PR #203, opened 2024-10-07
Bugfix: missing colon in `dz_m` in CAMx reader

blychs:
The colon after the height variable was missing, causing a mistake in the subtraction.
Trivial fix for that.


blychs:
@zmoon

zmoon (approved):

PR #205, opened 2024-10-08
Add unit conversion, variable load limit to RAQMS

### LINKED ISSUE #134: Unit conversion for gases in RAQMS reader?
@mebruckner expressed interest in this

mbruckner-work:
Additions to RAQMS reader:

1) Convert gases to ppbv ; resolves #134 
2) option for loading a small set of variables (consistent with _mm readers)

Before merging plan to add renaming or calculation for temperature_k and dz_m (needed for some aircraft and satellite pairings). These updates will be post-tutorial.

### DIFF
--- a/monetio/models/raqms.py
+++ b/monetio/models/raqms.py
@@ -35,7 +35,13 @@ def open_dataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
     return ds
 
 
-def open_mfdataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
+def open_mfdataset(
+    fname, 
+    convert_to_ppb=True, 
+    var_list=None,
+    surf_only=False, 
+    **kwargs
+):
     """Open a multiple file dataset from RAQMS output.
 
     Parameters
@@ -44,6 +50,10 @@ def open_mfdataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
         Files to be opened, expressed as a glob string or list of string paths.
     convert_to_ppb : boolean
         If true the units of the gas species will be converted to ppbv
+    var_list: list
+        List of variables to include in output. MELODIES-MONET only reads in
+        variables need to plot in order to save on memory and simulation cost
+        especially for vertical data. If None, will read in all model data.
         
     Returns
     -------
@@ -56,8 +66,11 @@ def open_mfdataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
             "in netCDF format."
             "Do not mix and match file types."
         )
-
-    ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")
+    if var_list is not None:
+        var_list.extend(['lat','lon',"IDATE", "Times","psfc","delp",'pdash'])
+        ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")[var_list]
+    else:
+        ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")
     ds = _fix(ds, surf_only=surf_only)
     
     # convert all gas species to ppbv
--- a/monetio/models/raqms.py
+++ b/monetio/models/raqms.py
@@ -51,7 +51,7 @@ def open_mfdataset(
     convert_to_ppb : boolean
         If true the units of the gas species will be converted to ppbv
     var_list: list
-        List of variables to include in output. MELODIES-MONET only reads in
+        List of variables to include in output. MELODIES-MONET should only read in
         variables need to plot in order to save on memory and simulation cost
         especially for vertical data. If None, will read in all model data.
         
--- a/monetio/models/raqms.py
+++ b/monetio/models/raqms.py
@@ -9,14 +9,16 @@ More information: http://raqms-ops.ssec.wisc.edu/
 import xarray as xr
 
 
-def open_dataset(fname, *, surf_only=False):
+def open_dataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
     """Open a single dataset from RAQMS output. Currently expects netCDF file format.
 
     Parameters
     ----------
     fname : str
         File to be opened.
-
+    convert_to_ppb : boolean
+        If true the units of the gas species will be converted to ppbv
+        
     Returns
     -------
     xarray.Dataset
@@ -33,14 +35,16 @@ def open_dataset(fname, *, surf_only=False):
     return ds
 
 
-def open_mfdataset(fname, *, surf_only=False):
+def open_mfdataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
     """Open a multiple file dataset from RAQMS output.
 
     Parameters
     ----------
     fname : str or list of str
         Files to be opened, expressed as a glob string or list of string paths.
-
+    convert_to_ppb : boolean
+        If true the units of the gas species will be converted to ppbv
+        
     Returns
     -------
     xarray.Dataset
@@ -55,6 +59,14 @@ def open_mfdataset(fname, *, surf_only=False):
 
     ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")
     ds = _fix(ds, surf_only=surf_only)
+    
+    # convert all gas species to ppbv
+    if convert_to_ppb:
+        for i in ds.variables:
+            if "units" in ds[i].attrs:
+                if "ppv" in ds[i].attrs["units"]:
+                    ds[i] *= 1e9
+                    ds[i].attrs["units"] = "ppbv"
 
     return ds
 
--- a/monetio/models/raqms.py
+++ b/monetio/models/raqms.py
@@ -32,6 +32,14 @@ def open_dataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
     ds = xr.open_dataset(names[0], drop_variables=["theta"])
     ds = _fix(ds, surf_only=surf_only)
 
+    # convert all gas species to ppbv
+    if convert_to_ppb:
+        for i in ds.variables:
+            if "units" in ds[i].attrs:
+                if "ppv" in ds[i].attrs["units"]:
+                    ds[i] *= 1e9
+                    ds[i].attrs["units"] = "ppbv"
+    
     return ds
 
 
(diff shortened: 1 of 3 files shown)


monetio/models/raqms.py
```diff
@@ -30,17 +32,37 @@ def open_dataset(fname, *, surf_only=False):
...
+    **kwargs
+):
     """Open a multiple file dataset from RAQMS output.
 
     Parameters
     ----------
     fname : str or list of str
         Files to be opened, expressed as a glob string or list of string paths.
-
+    convert_to_ppb : boolean
```
zmoon:
```suggestion
    convert_to_ppb : bool
```

monetio/models/raqms.py
```diff
@@ -30,17 +32,37 @@ def open_dataset(fname, *, surf_only=False):
     ds = xr.open_dataset(names[0], drop_variables=["theta"])
     ds = _fix(ds, surf_only=surf_only)
 
+    # convert all gas species to ppbv
+    if convert_to_ppb:
+        for i in ds.variables:
+            if "units" in ds[i].attrs:
+                if "ppv" in ds[i].attrs["units"]:
```
zmoon:
```suggestion
                if ds[i].attrs["units"].startswith("ppv"):
```
a little more specific? (assuming I'm correct it would be in the start) or if it's always just 'ppv' you could use `== "ppv"` instead.

monetio/models/raqms.py
```diff
@@ -9,14 +9,16 @@
...
-def open_dataset(fname, *, surf_only=False):
+def open_dataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
     """Open a single dataset from RAQMS output. Currently expects netCDF file format.
 
     Parameters
     ----------
     fname : str
         File to be opened.
-
+    convert_to_ppb : boolean
```
zmoon:
```suggestion
    convert_to_ppb : bool
```

monetio/models/raqms.py
```diff
@@ -30,17 +32,37 @@ def open_dataset(fname, *, surf_only=False):
...
     """Open a multiple file dataset from RAQMS output.
 
     Parameters
     ----------
     fname : str or list of str
         Files to be opened, expressed as a glob string or list of string paths.
-
+    convert_to_ppb : boolean
+        If true the units of the gas species will be converted to ppbv
+    var_list: list
```
zmoon:
```suggestion
    var_list : list of str, optional
```

monetio/models/raqms.py
```diff
@@ -52,10 +74,27 @@ def open_mfdataset(fname, *, surf_only=False):
...
+            if "units" in ds[i].attrs:
+                if "ppv" in ds[i].attrs["units"]:
+                    ds[i] *= 1e9
+                    ds[i].attrs["units"] = "ppbv"
+    
+    if 'ttheta' in ds.keys():
+        # Calculate temperature from potential temperature
+        k = 0.28571428571428564 # metpy.constants kappa
+        ds['temperature_k'] = ds['ttheta']*(ds['pres_pa_mid']/100000)**k
+        ds['temperature_k'].attrs['units'] = 'K'
```
zmoon: Do you want to add `long_name` as well?

monetio/models/raqms.py
```diff
@@ -52,10 +74,27 @@ def open_mfdataset(fname, *, surf_only=False):
             "in netCDF format."
             "Do not mix and match file types."
         )
-
-    ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")
+    if var_list is not None:
+        var_list.extend(['lat','lon',"IDATE", "Times","psfc","delp",'pdash','ttheta'])
+        ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")[var_list]
+    else:
+        ds = xr.open_mfdataset(names, concat_dim="time", drop_variables=["theta"], combine="nested")
```
zmoon: Most of these two lines are the same, so you could move it above the `if` and just do `ds = ds[var_list]` inside the `if`.

monetio/models/raqms.py
```diff
@@ -30,17 +32,37 @@ def open_dataset(fname, *, surf_only=False):
     ds = xr.open_dataset(names[0], drop_variables=["theta"])
     ds = _fix(ds, surf_only=surf_only)
 
+    # convert all gas species to ppbv
```
zmoon: You could move this to `_fix` or a `_fix_units` called by `_fix()` to you don't have to duplicate it?

monetio/models/raqms.py
```diff
@@ -9,14 +9,16 @@
 import xarray as xr
 
 
-def open_dataset(fname, *, surf_only=False):
+def open_dataset(fname, convert_to_ppb=True, surf_only=False, **kwargs):
```
zmoon: Currently `kwargs` are unused. Did you want to pass them to something?
```suggestion
def open_dataset(fname, convert_to_ppb=True, surf_only=False):
```

monetio/models/raqms.py
```diff
@@ -30,17 +32,37 @@ def open_dataset(fname, *, surf_only=False):
...
     ----------
     fname : str or list of str
         Files to be opened, expressed as a glob string or list of string paths.
-
+    convert_to_ppb : boolean
+        If true the units of the gas species will be converted to ppbv
+    var_list: list
+        List of variables to include in output. MELODIES-MONET should only read in
+        variables need to plot in order to save on memory and simulation cost
+        especially for vertical data. If None, will read in all model data.
```
zmoon:
```suggestion
        List of variables to include in output. MELODIES MONET should only read in
        variables needed to plot in order to save on memory and simulation cost
        especially for vertical data. If ``None`` (default), will read in all model data.
```

monetio/models/raqms.py
```diff
@@ -52,10 +74,27 @@ def open_mfdataset(fname, *, surf_only=False):
...
+    # convert all gas species to ppbv
+    if convert_to_ppb:
+        for i in ds.variables:
+            if "units" in ds[i].attrs:
+                if "ppv" in ds[i].attrs["units"]:
+                    ds[i] *= 1e9
+                    ds[i].attrs["units"] = "ppbv"
+    
+    if 'ttheta' in ds.keys():
+        # Calculate temperature from potential temperature
```
zmoon: Should `open_dataset` have this too?

monetio/models/raqms.py
```diff
@@ -52,10 +74,27 @@ def open_mfdataset(fname, *, surf_only=False):
...
+        k = 0.28571428571428564 # metpy.constants kappa
```
zmoon:
```suggestion
        k = 0.28571428571428564  # R/cp = kappa (unitless; value for dry air from metpy.constants)
```

monetio/models/raqms.py
zmoon:
```suggestion
```
I think your code is self-documenting enough, don't need this comment

zmoon (commented):
Added a few comments (and the format check has a few too). And would you mind updating the tests, at least to check for the unit conversion?

monetio/models/raqms.py
mbruckner-work: I was following the formatting used by the other readers. Switching to `== "ppv"`.

monetio/models/raqms.py
mbruckner-work: Moved into _fix, along with the unit conversion. Should now be applied to both.

monetio/models/raqms.py
```diff
@@ -9,13 +9,15 @@
 import xarray as xr
 
 
-def open_dataset(fname, *, surf_only=False):
+def open_dataset(fname, convert_to_ppb=True, surf_only=False):
```
zmoon:
```suggestion
def open_dataset(fname, *, convert_to_ppb=True, surf_only=False):
```
I think it makes sense for everything except `fname` to be keyword-only

monetio/models/raqms.py
```diff
@@ -28,18 +30,24 @@ def open_dataset(fname, *, surf_only=False):
...
 
     ds = xr.open_dataset(names[0], drop_variables=["theta"])
-    ds = _fix(ds, surf_only=surf_only)
+    ds = _fix(ds, surf_only=surf_only, convert_to_ppb=convert_to_ppb)
 
     return ds
 
 
-def open_mfdataset(fname, *, surf_only=False):
+def open_mfdataset(fname, convert_to_ppb=True, var_list=None, surf_only=False):
```
zmoon:
```suggestion
def open_mfdataset(fname, *, convert_to_ppb=True, var_list=None, surf_only=False):
```

monetio/models/raqms.py
```diff
@@ -28,18 +30,24 @@ def open_dataset(fname, *, surf_only=False):
...
 
     Parameters
     ----------
     fname : str or list of str
         Files to be opened, expressed as a glob string or list of string paths.
+    convert_to_ppb : bool
+        If true the units of the gas species will be converted to ppbv
+    var_list : list of str, optional
+        List of variables to include in output. MELODIES MONET should only read in
+        variables need to plot in order to save on memory and simulation cost
```
zmoon:
```suggestion
        variables needed to plot in order to save on memory and simulation cost
```

monetio/models/raqms.py
zmoon: @mbruckner-work if you don't want this, let me know and/or mark resolved and I'll approve the PR.

monetio/models/raqms.py
```diff
@@ -52,21 +60,36 @@ def open_mfdataset(fname, *, surf_only=False):
...
     ds = _fix_pres(ds)
 
     if surf_only:
         ds = ds.isel(z=0).expand_dims("z")
 
+    if convert_to_ppb:
+        for i in ds.variables:
+            if "units" in ds[i].attrs:
+                if ds[i].attrs["units"] == "ppv":
+                    ds[i] *= 1e9
```
zmoon: This seems to be causing an issue. In CI got `FAILED tests/test_raqms.py::test_surf_only[open_dataset] - ValueError: output array is read-only` for this line.

We could try it this way instead:
```suggestion
                    with xr.set_options(keep_attrs=True):
                        ds[i] = ds[i] * 1e9
```

mbruckner-work: Good to know that about CI. Will keep that in mind for future commits.

zmoon: It worked the original way on my machine. And in that run in CI it worked in Python 3.9 but failed in 3.10 and 3.11. I looked a bit at the env lists in the runs but nothing jumped out at me so far as to why that would happen. Seems like it would be related to `numpy` or `netcdf4` or `xarray` version but those were the same in 3.9 and 3.10.

zmoon (approved):
Thanks @mbruckner-work !

PR #164, opened 2024-03-12
Add Gridded VIIRS from the NOAA AWS bucket and more

bbakernoaa:
Added a new function to stream data from AWS from the noaa-jpss bucket into xarray for monetio.  This should be useful for satellite comparisons.  

It can handle both SNPP and NOAA20 satellites as well as varying resolutions (0.05, 0.1, 0.25 degree) and different time aggregations that are preprocessed by NESDIS (daily, weekly, monthly).  

Data availability is from 2012 -> current. 

to use:

```python 
nests_viirs_aod_gridded.open_dataset('2013-01-01', 'SNPP', '0.100', 'daily')

nests_viirs_aod_gridded.open_mfdataset(dates, 'SNPP', '0.050', 'daily')

``` 

@angehung5

PR #209, opened 2024-10-16
Develop VIIRS

dwfncar:
Added VIIRS L2 reader, based on MODIS L2.


bbakernoaa:
Also see https://github.com/noaa-oar-arl/monetio/pull/164

PR #214, opened 2024-11-05
Tropomi NO2 reader update

mbruckner-work:
Initial TROPOMI reader only worked with the reprocessed dataset (filename structure: "S5P_RPRO_L2__NO2____*"). 
This pull request includes the small change needed for the reader to work with the offline version (filename structure: "S5P_OFFL_L2__NO2____").

### DIFF
--- a/monetio/sat/_tropomi_l2_no2_mm.py
+++ b/monetio/sat/_tropomi_l2_no2_mm.py
@@ -103,8 +103,9 @@ def _open_one_dataset(fname, variable_dict):
             # Tropopause pressure
             ptrop = xr.full_like(itrop, np.nan, dtype=ds["preslev"].dtype)
             for i in np.unique(itrop):
-                if i >= 0 and i < p.sizes["z"]:
-                    ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
+                if np.isnan(i) or i < 0 or i >= p.sizes["z"]:
+                    continue
+                ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
             ds["troppres"] = ptrop
             ds["troppres"].attrs.update({"long_name": "tropopause pressure", "units": "Pa"})
 
--- a/monetio/sat/_tropomi_l2_no2_mm.py
+++ b/monetio/sat/_tropomi_l2_no2_mm.py
@@ -103,9 +103,8 @@ def _open_one_dataset(fname, variable_dict):
             # Tropopause pressure
             ptrop = xr.full_like(itrop, np.nan, dtype=ds["preslev"].dtype)
             for i in np.unique(itrop):
-                if np.isnan(i):
-                    continue
-                ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
+                if i >= 0 and i < p.sizes["z"]:
+                    ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
             ds["troppres"] = ptrop
             ds["troppres"].attrs.update({"long_name": "tropopause pressure", "units": "Pa"})
 


monetio/sat/_tropomi_l2_no2_mm.py
```diff
@@ -103,9 +103,8 @@ def get_extra(varname_, *, dct_=None, default_group="PRODUCT"):
             # Tropopause pressure
             ptrop = xr.full_like(itrop, np.nan, dtype=ds["preslev"].dtype)
             for i in np.unique(itrop):
-                if np.isnan(i):
-                    continue
-                ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
+                if i >= 0 and i < p.sizes["z"]:
```
zmoon: When would `i` be outside this range? (besides NaN)

mbruckner-work: In the OFFL files there are cases where the "tm5_tropopause_layer_index" has values outside of this range, which causes a problem when those values are used to index into the pressure array to get the tropopause pressure.

zmoon: Just  $\ge$ `p.sizes["z"]`? or negative values too? and do you know why?

mbruckner-work: The tropopause layer index in the OFFL files I checked did have negative values in it.

monetio/sat/_tropomi_l2_no2_mm.py
```diff
@@ -103,9 +103,8 @@ def get_extra(varname_, *, dct_=None, default_group="PRODUCT"):
...
+                    ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
```
zmoon: How about this, to preserve the notion in the code that it can be NaN.
```suggestion
                if np.isnan(i) or i < 0 or i >= p.sizes["z"]:
                    continue
                ptrop = xr.where(itrop == i, p.isel(z=int(i)), ptrop)
```

mbruckner-work: The ptrop array is set up initially filled with NaN and then replaces those with a pressure where there is a valid tropopause index to use. 
Both are giving the same result, so whichever syntax is preferred is fine.

zmoon: > The ptrop array is set up initially filled with NaN and then replaces those with a pressure where there is a valid tropopause index to use.

Yes, thanks, I wrote this part 😅

> whichever syntax is preferred is fine

I prefer preserving the notion that NaN can be one of the unique trop index values.

mbruckner-work: Okay- there's just no actual difference in the output because NaN is getting replaced with NaN for a trop index of NaN.

zmoon: > no actual difference in the output

I know. Like I've said, I'd just prefer to preserve the notion in the code that NaN can be one of the unique trop index values.

>  NaN is getting replaced with NaN for a trop index of NaN

No it's not, it's being skipped since `p.isel(z=int(i))` would fail. `i >= 0 and i < p.sizes["z"]`, what you used, also guards against NaN `i`, it's just not explicit about it.

zmoon (approved):
Thanks @mbruckner-work

monetio/sat/_tropomi_l2_no2_mm.py
mbruckner-work: > No it's not, it's being skipped since `p.isel(z=int(i))` would fail. `i >= 0 and i < p.sizes["z"]`, what you used, also guards against NaN `i`, it's just not explicit about it.

Whoops- I mis-read what was getting past the check and when the where gets evaluated. I'm not very familiar with using 'continue'.

PR #217, opened 2024-11-14
Add model reader for gocart output

MaggieMarvin:
These changes include the addition of a model reader for gocart output that can be used to process GEFS aerosol files. As written, this code reads in total AOD at 4 wavelengths (470, 550, 670, and 870 nm) and currently drops/ignores the speciated AOD variables generated by gocart (AOD_BC, AOD_DU, AOD_OC, AOD_SS, and AOD_SU), but could be extended to include those variables if they're likely to be used regularly.


zmoon:
@MaggieMarvin do you have an example file or two so we can write a test?

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
...
+        ("y", "x"),
+        lat,
+        {
+            "long_name": "Latitude",
+            "units": "degree_north",
+            "standard_name": "latitude",
+        },
+    )
+    ds = ds.reset_coords().set_coords(["time", "latitude", "longitude"])
+    del lon, lat
```
zmoon:
```suggestion

```
they will get gc'ed automatically, being scoped to this function

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def open_mfdataset(fnames):
+    """Method to open GOCART netcdf files.
+
+    Parameters
+    ----------
+    fname : string or list
+        fname is the path to the file or files.  It will accept hot keys in
+        strings as well.
```
zmoon:
```suggestion
    fnames
        String glob expression or a list of files to open.
```

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Returns
+    -------
+    xarray.DataSet
```
zmoon:
```suggestion
    xarray.Dataset
```

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
...
+        GOCART model dataset in standard format for use in MELODIES-MONET
```
zmoon:
```suggestion
        GOCART model dataset in standard format for use in MELODIES MONET
```

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
...
+
+    """
+
+    # open the dataset using xarray
```
zmoon:
```suggestion
```

monetio/models/gocart.py
```diff
@@ -0,0 +1,78 @@
+""" GOCART File Reader """
```
zmoon:
```suggestion
"""GOCART File Reader."""
```
Also, could you add a bit more info about GOCART (maybe a link or two) and contexts that this reader is applicable to? Is it only for GEFS-Aerosols or is it applicable to other GOCART situations?

zmoon (commented):
@MaggieMarvin this is looking good so far, have you tested it in MM already?

MaggieMarvin:
@zmoon Great, thanks! Yes I've tested in melodies-monet - the model files I used are here on Hera: /scratch2/NCEPDEV/stmp3/Margaret.R.Marvin/global/stmp/RUNDIRS/WetDep_Dust/gefsfcst.2018081600/fcst.371442/gocart.inst_aod*.nc4

monetio/models/gocart.py
MaggieMarvin: Still looking into this - I'll get back to you soon

monetio/models/gocart.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+
+    ds = xr.open_mfdataset(
+        fnames,
+        concat_dim="time",
+        combine="nested",
+        drop_variables=["AOD_BC", "AOD_DU", "AOD_OC", "AOD_SS", "AOD_SU"],
+    )
+
+    ds["AOD470"] = ds.AOD.isel(lev=0)
```
zmoon: Could you set it up so the user can specify the wavelengths (in case they are different)? It can default to these four though.

monetio/models/gocart.py
```diff
@@ -0,0 +1,75 @@
...
+    xarray.Dataset
+        GOCART model dataset in standard format for use in MELODIES MONET
+
+    """
+
+    ds = xr.open_mfdataset(
+        fnames,
+        concat_dim="time",
+        combine="nested",
+        drop_variables=["AOD_BC", "AOD_DU", "AOD_OC", "AOD_SS", "AOD_SU"],
```
zmoon: User might want these. If so, I guess you'd want to select and rename them too.
```suggestion
```

monetio/models/gocart.py
```diff
@@ -0,0 +1,75 @@
+"""GOCART File Reader."""
+
+import xarray as xr
+
+
+def open_mfdataset(fnames):
```
zmoon:
```suggestion
def open_mfdataset(fnames, **kwargs):
```
Can pass `**kwargs` to `xr.open_mfdataset()` (e.g. to drop those variables you didn't want).

monetio/models/gocart.py
MaggieMarvin: Sure! I hadn't thought of a good naming convention since these variables are also wavelength-dependent - for example could do "AOD550_DU" or something else? Unless wavelength is a required argument as above, then they can all keep their original names

monetio/models/gocart.py
MaggieMarvin: I believe it's only these four wavelengths that are available from gocart - should wavelength be a required argument and then we can pass all AOD variables at that wavelength? Error if a wavelength is requested that isn't provided?

zmoon: Barry told me the wavelengths are technically configurable. My thought was the wavelength list could be an optional argument (defaulting to your four if not provided), but then we make sure the list length is the same as the `lev` dim size.

zmoon:
I extracted a small spatial subset of one of the files to use for test. [gocart.inst_aod.20180901_1200z.subset.nc4.zip](https://github.com/user-attachments/files/17853150/gocart.inst_aod.20180901_1200z.subset.nc4.zip)

monetio/models/gocart.py
zmoon: Maybe just always stick the wavelength on the end of the initial name?

zmoon:
Still a bit to do here.

PR #222, opened 2025-01-30
Address codespell updates

zmoon:
It found a new misspelling but also now thinks "indx" should be "index"


zmoon:
Some ISH tests are flaking

Issue #198, opened 2024-10-03
OpenAQ web API v3


zmoon:
This was announced today: https://openaq.medium.com/announcing-openaq-version-3-api-5d67fe3b7a3a

They also said they currently plan "deprecation of version 1 and version 2 APIs effective January 31, 2025". So we should create a new v3 reader based on the current one for v2 relatively soon.

zmoon:
The OpenAQ v2 tests failed in the scheduled CI run this morning with HTTP error 410 ("Gone").

PR #224, opened 2025-02-07
xfail OpenAQ web API v2 tests

zmoon:
See #198


zmoon:
The GML ozonesonde pages are not down but seem to be slow to respond currently.

PR #229, opened 2025-02-14
Simple bugfix: PM2.5 should be dry

blychs:
Measurements are always in dry PM2.5, the current code used wet.
This PR fixes that by removing a single line of code, which included PH2O in the particulate matter.
Cheers
Pablo


zmoon (approved):
👍

zmoon:
Would've merged this already but I was trying to see if we could get ✅. But it seems like the 3.8 case is having trouble solving the Conda env. I will work on that separately, since it's not really related to this PR.

PR #232, opened 2025-02-25
Use archive link for AirNow description attribution

zmoon:
alerted to this because the link returned 404 in #228, but it seems like a while before this (at least as long ago as fall 2023, from the web archive) it just redirected to the home page


zmoon:
currently https://www.epa.gov/castnet is down

PR #234, opened 2025-02-26
Remove later-renamed vars from var list in RRFS-CMAQ MM reader

### LINKED ISSUE #233: UFS-AQM reader temperature pairing
The UFS-AQM (rrrfs) reader errors when trying to pair on temperature. I think pressure would have the same problem. Basically this is because we rename pressure and temperature to standard names {"temperature_k", "pres_pa_mid"} when reading these inputs in. 

It is the same problem as that fixed here in the WRF-Chem reader. So something like line 100 in WRF-Chem needs to be added to this reader too: https://github.com/noaa-oar-arl/monetio/blob/develop/monetio/models/_wrfchem_mm.py

I'm not sure the best way to do this, but I think you need to remove the standard temperature and pressure names from the var_list like we do with the PM2.5 products on line 120. So if you include these in the mapping since they are automatically already included and then renamed and no need to include them in the var_list.

zmoon:
Resolves #233

### DIFF
--- a/monetio/models/_rrfs_cmaq_mm.py
+++ b/monetio/models/_rrfs_cmaq_mm.py
@@ -111,6 +111,19 @@ def open_mfdataset(
         var_list.append("hgtsfc")
         var_list.append("delz")
 
+        # Remove standardized variable names that user may have requested to pair on in MM
+        # as they will be added anyway and here would cause `[var_list]` to fail.
+        for vn in [
+            "temperature_k",
+            "surfpres_pa",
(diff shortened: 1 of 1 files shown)


zmoon:
@rschwant can you check if this resolves your issue?

rschwant:
Yes, this fixes the problem exactly and now we can use this tool to make plots to evaluate temperature! I tested it and now get plots and it did not break any of our other plots either.

PR #235, opened 2025-02-28
OpenAQ web API v3

### LINKED ISSUE #198: OpenAQ web API v3

zmoon:
Resolves #198 

Some changes from v2 to v3:
* Have to request data from each sensor separately
  - each sensor has data for one parameter (ID) only
  - to facilitate selecting by country, e.g., have to know all the available locations/sensors, which takes a minute or two to get (implemented caching to help)
* Some differences in the behavior wrt. specifying time period in the request
* No more search radius
* No more sensor type (low-cost vs reference-grade), just the `isMonitor` bool
* hourly- and daily-average data, in addition to raw
* ...

Note: `filelock` should be moved to the optional dependencies section of the Conda env spec and included in the conda-forge recipe.

cc: @jordanschnell

### DIFF
--- a/monetio/obs/openaq_v3.py
+++ b/monetio/obs/openaq_v3.py
@@ -1,0 +1,611 @@
+"""Get AQ data from the OpenAQ v2 REST API.
+
+Visit https://docs.openaq.org/docs/getting-started to get an API key
+and set environment variable ``OPENAQ_API_KEY`` to use it.
+
+For example, in Bash:
+
+.. code-block:: bash
+
+   export OPENAQ_API_KEY="your_api_key_here"
+
+https://openaq.org/
+
+https://api.openaq.org/docs#/v2
+"""
+
+import functools
+import json
+import logging
+import os
+import warnings
+
+import numpy as np
+import pandas as pd
+import requests
+
+logger = logging.getLogger(__name__)
+
+API_KEY = os.environ.get("OPENAQ_API_KEY", None)
+if API_KEY is not None:
+    API_KEY = API_KEY.strip()
+    if len(API_KEY) != 64:
+        warnings.warn(f"API key length is {len(API_KEY)}, expected 64")
+
+_PPM_TO_UGM3 = {
+    "o3": 1990,
+    "co": 1160,
+    "no2": 1900,
+    "no": 1240,
+    "so2": 2650,
+    "ch4": 664,
+    "co2": 1820,
+}
+"""Conversion factors from ppmv to µg/m³.
+
+Based on
+
+- air average molecular weight: 29 g/mol
+- air density: 1.2 kg m -3
+
+and rounded to 3 significant figures.
+"""
+
+# NOx assumption
+_PPM_TO_UGM3["nox"] = _PPM_TO_UGM3["no2"]
+
+_NON_MOLEC_PARAMS = [
+    "pm1",
+    "pm25",
+    "pm4",
+    "pm10",
+    "bc",
+]
+"""Parameters that are not molecules and should be in µg/m³ units."""
+
+
+def _api_key_warning(func):
+    @functools.wraps(func)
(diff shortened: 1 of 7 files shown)


zmoon:
Tests failing are ICAP-MME, SSL cert out of date

monetio/obs/openaq_v3.py
```diff
@@ -0,0 +1,855 @@
...
+    if sensor_type is not None:
+        # FIXME: may not be the best approach
+        meta = meta.assign(
+            sensor_type=meta["is_monitor"].map(
+                {
+                    True: "reference grade",
+                    False: "low-cost sensor",
+                }
+            )
+        )
```
zmoon: @russbiggs is the isMonitor true/false breakdown the same as the old "reference grade" vs "low-cost sensor"?

russbiggs: In v3 it's `monitor`. So `?monitor=true` will return just reference monitors, `?monitor=false` will return just air sensors (aka low-cost sensor). Not including this parameter will include both types.

zmoon: Thanks, seems like my approach here is correct then.

monetio/obs/openaq_v3.py
```diff
@@ -0,0 +1,855 @@
...
+            r = requests.get(url, params=params, headers=headers, timeout=timeout)
+            tries += 1
+            if r.status_code == 408:
+                logger.info(f"request timed out (try {tries}/{retry})")
+                time.sleep(tries + 0.1 * rand())
+            elif r.status_code == 429:
+                # Note: response headers don't seem to include Retry-After
+                # Just `{'detail': 'To many requests'}` in `.json()`
+                logger.info(f"rate limited (try {tries}/{retry})")
+                time.sleep(tries * 5 + 0.2 * rand())
```
zmoon: @russbiggs do you have any comments about this approach? Would you recommend backing off more initially when rate-limited? Is there some way to get Retry-After? Or is it always the same?

russbiggs: The preferred approach is to use the providers rate limit headers in the HTTP response which you can read about here:

https://docs.openaq.org/using-the-api/rate-limits#rate-limit-headers

In the official OpenAQ python SDK you can see how i'm exposing them, (I'm using httpx but its pretty close to requests).

https://github.com/openaq/openaq-python/blob/main/openaq/shared/responses.py#L147

If you bundle these values with the response (or they would already be there if using the response object directly) then you can access and throttle before the next request something like:

```py
if req.headers["x-ratelimit-remaining"] < 1:
    time.sleep(req.headers["x-ratelimit-reset"])
# else proceed to next request
```

Then you will sleep for exactly long as is needed until the rate limit resets and you'll get a fresh allocation of requests, no need to exponential back off or guessing.

zmoon: Gotcha, I'll add that possible wait for the 200 case, I can see those x-ratelimit response headers are there. However, for the 429 case, which can still happen with multi-threaded requesting, it seems those headers are not currently included.

russbiggs: Good catch, if you submit a ticket here https://github.com/openaq/openaq-api-v2 i dont see why they couldnt be added. i think it would be a relatively straightforward add

monetio/obs/openaq_v3.py
zmoon: `is_monitor` that I'm using in this code block is `isMonitor` from the locations endpoint data. I assume the `monitor` query parameter you mention is consistent with that.

bbakernoaa (approved):

Issue #242, opened 2025-05-09
Optimization of MM efficiency


angehung5:
We are using [MELODIES-MONET](https://github.com/NCAR/MELODIES-MONET) for AQMv8_p1 evaluation, to generate paired netcdf files and plots such as timeseries and taylor diagram However, MM reads the entire 3D/4D variable from AQM outputs, it requires huge disk memory for 1 month of simulation and usually leads to OUT OF MEMORY errors. An additional step to extract the surface fields would be required, which we want to avoid. Have discussed with @bbakernoaa, it should be able to add an option to let MM read the first layer only.

Also MM requires users manually update control yaml files with desired dates and species. A more automaitc workflow would be good, such as making control yaml updated depending on the main python script, so users don't need to manually change every yaml files.

@bbakernoaa has created a MM module on Gaea C6:

/gpfs/f6/bil-fire3/world-shared/modulefiles/melodies-monet/

and here are the MM scripts we are using:

/gpfs/f6/bil-fire3/world-shared/Wei-ting.Hung/aqmv8p1_eval/MM_scripts

@benkozi @iri01

benkozi:
I think this feature would give me a good opportunity to familiarize myself with the library. I'll plan to take a look early next week and report back.

bbakernoaa:
I think that the reader itself already has a surface only option. It may be
that we missed a setting in MM or need to add that as an option in the MM
driver.


Barry Baker

National Oceanic and Atmospheric Administration
Air Resources Laboratory
Physical Research Scientist
Chemical Modeling and Emissions Group Leader
NCWCP, R/ARL, Rm. 4204
5830 University Research Court
College Park, Maryland  20740
Phone: ‪(301) 683-1395‬

benkozi:
@bbakernoaa - Is this the reader for `mod_type: "rrfs"`? https://github.com/noaa-oar-arl/monetio/blob/stable/monetio/models/_rrfs_cmaq_mm.py#L16.

bbakernoaa:
thats correct.  It should simply be named ufs.py.  There was a split that needs to be corrected in that specialized readers were made for MM when there shouldn't be.

benkozi:
@bbakernoaa - Thanks. I can change the name as part of the work.

~~I'm not seeing a surface only option in the reader. I'll look into adding one. I'm not sure how lazy loading and `xarray.open_mfdataset` interact especially with slicing, so that's where it might get a bit complicated.~~

[...]

PR #243, opened 2025-05-12
[develop] feat:  add test for surf_only in UFS-AQM driver

benkozi:
* Adds a test for `surf_only` ensuring that data is lazy loaded for most data variables following from discussion in #242.
* Renames "rrfs-cmaq" driver to `ufs.py`. This change will likely affect upstream driver mapping.


monetio/models/__init__.py
```diff
@@ -15,14 +14,14 @@
     pardump,
     prepchem,
     raqms,
+    ufs,
```
benkozi, line 18: Could you point me to how I make this change backwards compatible with `obs_model: "rrfs"`?

cc @bbakernoaa @zmoon

benkozi, line 18: I had intended to write documentation, but imo it makes sense to have a more involved doc update including and related to @angehung5's request to add control YAML updating from the main script:

> Also MM requires users manually update control yaml files with desired dates and species. A more automaitc workflow would be good, such as making control yaml updated depending on the main python script, so users don't need to manually change every yaml files.

I think this request should be a separate issue (which I could implement). What would you recommend?

cc @bbakernoaa @zmoon

zmoon: For backwards compat here, one approach would be to recreate the old module with functions (e.g. `open_mfdataset`) that wrap (`functools.wraps`) the ones from `ufs.py`, and have the wrappers raise `DeprecationWarning`s to suggest moving to the new module.

environment-dev.yml
```diff
@@ -25,9 +25,11 @@ dependencies:
   - requests
   - timezonefinder
   #
-  # test
+  # test/dev
   - pytest
   - pytest-xdist
+  - pytest-mock
```
zmoon: We try for alphabetical order in sections
```suggestion
  - pytest-mock
  - pytest-xdist
```

environment-dev.yml
```diff
@@ -25,9 +25,11 @@ dependencies:
...
+  - pre-commit
```
zmoon: `pre-commit` works well as a globally-installed tool (e.g. with `pipx install` or `uv tool install`). I like to support that pattern (though of course one can add it to the conda env instead after creating it if they desire). Open to thoughts though.
```suggestion
```

tests/test_ufs.py
```diff
@@ -0,0 +1,46 @@
...
+)
+def test_open_mfdataset_surf_only(
+    data_dir: Path, test_data: SurfOnlyTestData, mocker: MockerFixture
+) -> None:
+    slug = "aqm.t12z.dyn.f*.nc"
+    spy = mocker.spy(monetio.models.ufs, "_isel_surface_level_")
+    actual = open_mfdataset(str(data_dir / "ufs" / slug), surf_only=test_data.surf_only)
+
+    # Confirm the expeced function is called...perhaps overkill
+    assert spy.call_count == int(test_data.surf_only)
```
zmoon: I would agree that this is overkill, since you are checking the dim sizes.

tests/test_ufs.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for var in actual.data_vars.values():
+        shape_dict = {dim: actual.dims[dim] for dim in var.dims}
```
zmoon: Currently [`ds.dims`](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.dims.html) works for this, but it may not in the future (`ds.sizes` the replacement).
```suggestion
        shape_dict = {dim: actual.sizes[dim] for dim in var.dims}
```

zmoon:
The original rationale for making these modules "private" (names with leading underscore) was that we were taking them in without any unit tests. So this is nice. Thanks @benkozi

bbakernoaa:
@benkozi @zmoon 

we may also update the `_calc_pressure` and `_calc_hgt` functions to be vectorized rather than loops. We also are loading the data which puts all the data into memory 

```python
def _calc_pressure(self, dset):
        """Calculate the mid-layer pressure in Pa.

        Parameters
        ----------
        dset : xarray.Dataset
            The UFS dataset

        Returns
        -------
        xarray.DataArray
            Mid-layer pressure
        """
        # Create a copy of dp_pa array with the same dimensions
        p = dset.dp_pa.copy()

        # Get surface pressure for calculation
        psfc = dset.surfpres_pa

        # Calculate pressure at each layer boundary using vectorized operations
        # First, calculate pressure at layer interfaces using ak and bk coefficients
        p_interfaces_1 = dset.ak[:-1] + psfc * dset.bk[:-1]  # Upper interface of each layer
        p_interfaces_2 = dset.ak[1:] + psfc * dset.bk[1:]    # Lower interface of each layer

        # Calculate mid-layer pressure using log-average
        # log(p_mid) = (p_2 - p_1) / ln(p_2/p_1)
        # This preserves all dimensions and allows lazy evaluation
        p_mid = (p_interfaces_2 - p_interfaces_1) / np.log(p_interfaces_2 / p_interfaces_1)

        # Set attributes
        p_mid.name = "pres_pa_mid"
        p_mid.attrs["units"] = "pa"
        p_mid.attrs["long_name"] = "Pressure Mid Layer in Pa"

        return p_mid

    def _calc_hgt(self, dset):
        """Calculate the geopotential height in m.

        Parameters
        ----------
        dset : xarray.Dataset
            The UFS dataset

        Returns
        -------
        xarray.DataArray
            Geopotential height
        """
        # Get surface altitude
        sfc = dset.surfalt_m

        # Get the vertical displacement and flip sign as needed
        dz = dset.dz_m * -1.0

        # Add surface elevation to the bottom layer using vectorized operation
        # Create a template for the surface altitude with same shape as dz's horizontal dims
        template = dz.isel(z=0).copy()

        # Add surface elevation
        dz = dz.where(~(dz.z == dz.z[0]), dz + sfc)

        # Calculate cumulative sum along z dimension to get heights
        z = dz.cumsum(dim='z')

        # Set attributes
        z.name = "alt_msl_m_full"
        z.attrs["long_name"] = "Altitude MSL Full Layer in Meters"
        z.attrs["units"] = "m"

        return z
        ```

File: monetio/models/__init__.py (hunk 1 above)
Comment (zmoon) [reply monetio/models/__init__.py:18]: On the MM side, we would update MM to try to import from the new one first then fall back to the old module. And we *could* just do that, since the module here was "private".

Comment (benkozi) [reply monetio/models/__init__.py:18]: I'm okay with either approach of course. The second suggestion seems a bit cleaner and also gives an opportunity to update the model type to also include `ufs` in addition to `rrfs`. I did some searching but am having trouble tracking down where this code change would occur. Could you point me in the right direction?

File: environment-dev.yml (hunk 3 above)
Comment (benkozi) [reply environment-dev.yml]: I have a Windows box and tend to do all my development in local Linux Docker container, so it makes sense to include it for that reason. However, I can install `pre-commit` in the Docker recipe only and leave it off the environment YML. Let me know what you prefer!

File: environment-dev.yml (hunk 2 above)
Comment (benkozi) [reply environment-dev.yml]: Considering the mock testing is not needed. I removed this dependency. Resolving

File: environment-dev.yml (hunk 3 above)
Comment (benkozi) [reply environment-dev.yml]: Resolving. You're right that this can be used at the global level. No need to install in the docker environment.

File: tests/test_ufs.py (hunk 5 above)
Comment (benkozi) [reply tests/test_ufs.py]: Changed and resolving

**Comment by benkozi** [comment]
> we may also update the `_calc_pressure` and `_calc_hgt` functions to be vectorized rather than loops. We also are loading the data which puts all the data into memory

@bbakernoaa - Thanks for the snippets. I swapped out the current implementations for these with some minor adjustments to the  mid-layer pressure calculation. I compared against the previous "baseline" and the results were identical. I didn't add unit testing since the functions are private, but I can add them.

File: tests/test_ufs.py (hunk 4 above)
Comment (benkozi) [reply tests/test_ufs.py]: Removed and resolving

**Comment by bbakernoaa** [comment]
> > we may also update the `_calc_pressure` and `_calc_hgt` functions to be vectorized rather than loops. We also are loading the data which puts all the data into memory
> 
> @bbakernoaa - Thanks for the snippets. I swapped out the current implementations for these with some minor adjustments to the mid-layer pressure calculation. I compared against the previous "baseline" and the results were identical. I didn't add unit testing since the functions are private, but I can add them.

Awesome! This should drastically help improve the speed when evaluating a large dataset

File: monetio/models/__init__.py (hunk 1 above)
Comment (benkozi) [reply monetio/models/__init__.py:18]: > I did some searching but am having trouble tracking down where this code change would occur. Could you point me in the right direction?

Oops...I was looking in the wrong repo. Thought `monet` would be it, but it's https://github.com/NCAR/MELODIES-MONET. Let me see about getting a PR together over there.

File: monetio/models/ufs.py [hunk 6]
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
+    sfc = dset.surfalt_m
+
+    # Get the vertical displacement and flip sign as needed
+    dz = dset.dz_m * -1.0
+
+    # Add surface elevation
+    dz = dz.where(~(dz.z == dz.z[0]), dz + sfc)
+
+    # Calculate cumulative sum along z dimension to get heights
+    z = dz.cumsum(dim="z")
```
zmoon: Could this not just be
```python
z = dz.cumsum(dim="z") + sfc
```
(simpler, and, in my experience, better to avoid `.where()` when you can)

monetio/models/ufs.py
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
+
+    # Set attributes
     z.name = "alt_msl_m_full"
     z.attrs["long_name"] = "Altitude MSL Full Layer in Meters"
```
zmoon:
```suggestion
    z.attrs["long_name"] = "altitude above MSL at full layer tops"
```
?

monetio/models/ufs.py
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
-    dz = f.dz_m.load() * -1.0
-    # These are negative in RRFS-CMAQ, but you resorted and are adding from the surface,
-    # so make them positive.
-    dz[:, 0, :, :] = dz[:, 0, :, :] + sfc  # Add the surface altitude to the first model level only
-    z = dz.rolling(z=len(f.z), min_periods=1).sum()
+    # Get surface altitude
+    sfc = dset.surfalt_m
+
+    # Get the vertical displacement and flip sign as needed
+    dz = dset.dz_m * -1.0
```
zmoon:
```suggestion
    dz = dset.dz_m
```
It's already been fixed before we call this func.

We should probably also move the fixing it (L209 below) inside the first cond branch.
https://github.com/noaa-oar-arl/monetio/blob/bbc0d8218f7128ca5d015b9bd8337fa7b116922a/monetio/models/ufs.py#L205-L209

monetio/models/ufs.py
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
+    # Calculate mid-layer pressure using log-average
+    # log(p_mid) = (p_2 - p_1) / ln(p_2/p_1)
+    # This preserves all dimensions and allows lazy evaluation
+    p_mid = (p_interfaces_2 - p_interfaces_1) / np.log(p_interfaces_2 / p_interfaces_1)
+    p_mid = p_mid.transpose(*dset.pm25_ave.dims)
+
+    # Set attributes
+    p_mid.name = "pres_pa_mid"
+    p_mid.attrs["units"] = "pa"
+    p_mid.attrs["long_name"] = "Pressure Mid Layer in Pa"
```
zmoon:
```suggestion
    p_mid.attrs["long_name"] = "mid-layer pressure"
```

monetio/models/ufs.py
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
+
+    # Calculate mid-layer pressure using log-average
+    # log(p_mid) = (p_2 - p_1) / ln(p_2/p_1)
+    # This preserves all dimensions and allows lazy evaluation
+    p_mid = (p_interfaces_2 - p_interfaces_1) / np.log(p_interfaces_2 / p_interfaces_1)
+    p_mid = p_mid.transpose(*dset.pm25_ave.dims)
+
+    # Set attributes
+    p_mid.name = "pres_pa_mid"
+    p_mid.attrs["units"] = "pa"
```
zmoon:
```suggestion
    p_mid.attrs["units"] = "Pa"
```

monetio/models/ufs.py
```diff
@@ -1053,69 +1053,76 @@ def dict_species_sums(mech):
...
+    ak = xr.DataArray(dset.ak, dims="z")
+    bk = xr.DataArray(dset.bk, dims="z")
+    p_interfaces_1 = ak[:-1] + psfc * bk[:-1]  # Upper interface of each layer
+    p_interfaces_2 = ak[1:] + psfc * bk[1:]  # Lower interface of each layer
+
+    # Calculate mid-layer pressure using log-average
+    # log(p_mid) = (p_2 - p_1) / ln(p_2/p_1)
+    # This preserves all dimensions and allows lazy evaluation
+    p_mid = (p_interfaces_2 - p_interfaces_1) / np.log(p_interfaces_2 / p_interfaces_1)
+    p_mid = p_mid.transpose(*dset.pm25_ave.dims)
```
zmoon: `pm25_ave` might not be present / might be better to be explicit
```suggestion
    p_mid = p_mid.transpose("time", "z", "y", "x")
```

zmoon:
Not saying this should be addressed in this PR, but it seems like this `alt_msl_m_full` layer-top height isn't usually what people want. They usually want `alt_msl_m_mid` (mid-layer).

monetio/models/ufs.py
benkozi: Done. I had done this originally :-). Resolving

monetio/models/ufs.py
benkozi: Done. Resolving

monetio/models/ufs.py
benkozi: Done. Resolving

monetio/models/ufs.py
benkozi: > Get the vertical displacement and flip sign as needed
    dz = dset.dz_m * -1.0

Removing this line changed answers, and I'm not sure how it interacts with the suggested change about moving the in the main function. Moving the sign switch inside the conditional did not address the failure. I'm inclined to leave as is without additional testing, but let me know...

monetio/models/ufs.py
benkozi: Done. Resolving

monetio/models/ufs.py
benkozi: This change failed the baseline test. I expect it's because we're adding surface elevation to _everything_ including the surface. I agree that avoiding `where` is generally a good thing.

monetio/models/_rrfs_cmaq_mm.py
```diff
@@ -1,1121 +1,6 @@
...
-        pres_2 = dset.ak[k + 1] + psfc * dset.bk[k + 1]
-        pres_1 = dset.ak[k] + psfc * dset.bk[k]
-        p[:, k, :, :] = (pres_2 - pres_1) / np.log(pres_2 / pres_1)
-
-    p.name = "pres_pa_mid"
-    p.attrs["units"] = "pa"
-    p.attrs["long_name"] = "Pressure Mid Layer in Pa"
-    return p
+warnings.filterwarnings("default", category=DeprecationWarning, module=__name__)
+warnings.warn("_rrfs_cmaq_mm is deprecated. Use ufs instead.", DeprecationWarning)
```
benkozi: @zmoon - This warning is surfaced on import in my testing. It also allows all the old functions to still be imported.

monetio/models/ufs.py
zmoon: The "as needed" is the key there. It's not needed, since the sign has already been flipped before we call this function (although that should only be done in the case that the first condition in that block is true, not always done like it is now, which is a bug, which we can fix here).

I tested this branch, and I could see the heights were negative, which is not what we want (regardless of what it was before). Might as well fix it in this PR. Suggest adding to your test some assertions related to the height and pressure (min/max at expected ends of the _z_ dim, all positive, monotonicity, etc.).

monetio/models/ufs.py
zmoon: It seems like the develop branch result had sign issues, so maybe we shouldn't worry too much about comparing to it? Does my suggestion make sense to you otherwise? ~~`sfc` seems to be all zero in your test case, but~~ I was getting the same results.

monetio/models/_rrfs_cmaq_mm.py
zmoon: Yes but since we import this in the `models/__init__.py`, it will run every first import of `monetio`, no? (since we import from `models` into the top-level `__init__.py`) Not sure we want that.

monetio/models/ufs.py
```diff
@@ -1059,33 +1059,33 @@ def dict_species_sums(mech):
...
-    sfc = dset.surfalt_m
-
-    # Get the vertical displacement and flip sign as needed
+    # These are negative in UFS-AQM, but we resorted and are adding from the surface,
+    # so make them positive.
     dz = dset.dz_m * -1.0
 
-    # Add surface elevation
-    dz = dz.where(~(dz.z == dz.z[0]), dz + sfc)
+    # Add surface elevation to everything except surface
```
zmoon: I believe it's actually doing the opposite, only adding the surface elevation to the surface. `dz + dset.surfalt_m` (the *other*) is used where the cond is false. i.e. where `dz.z == dz.z[0]`

monetio/models/_rrfs_cmaq_mm.py
benkozi: That's a good point. Would you suggest we just wrap `open_mfdataset` then?

monetio/models/ufs.py
benkozi: Yeah, I noticed they were negative. I'll fix and update the baseline.

monetio/models/ufs.py
benkozi: > Does my suggestion make sense to you otherwise?

Sorry, do you mean avoiding the where?

monetio/models/_rrfs_cmaq_mm.py
zmoon: I think that would be the better solution. But I also think, given its "private" nature, we could get away with just removing the old module and not worrying about it, especially since you've already started the MM PR. 🤷‍♀️

benkozi: That was my original thought, but I started to get a little nervous about breaking upstream apps. I'll change to just wrap `open_mfdataset` which will be a good enough warning.

monetio/models/ufs.py
benkozi: I'll need to look more closely!

monetio/models/ufs.py
zmoon: Yes sorry. Using `z = dz.cumsum(dim="z") + dset.surfalt_m` instead the two lines that involve `.where()`. Results seem to be `np.isclose` the same, but not identical.

monetio/models/ufs.py
benkozi: I made the changes as suggested and updated the baseline. The new value description for `alt_msl_m_full`:
```
mean     15257.173828
std      13814.880859
min        785.901489
25%       3529.174255
50%      10933.270508
75%      23337.199707
max      60205.074219
Name: alt_msl_m_full, dtype: float64
```

monetio/models/_rrfs_cmaq_mm.py
benkozi: So, I decided to remove the override to always emit the import warning. It can be enabled at the global level by diligent software folks. I expect most users will come through MM. Also, IDEs will highlight the module as deprecated. It's not a huge deal to have a little phantom module around for awhile. I can get fancier if you'd like

monetio/models/ufs.py
benkozi: Resolving in favor of the other ongoing comment thread in the same vein

monetio/models/_rrfs_cmaq_mm.py
zmoon: Fine with me!

> I expect most users will come through MM

agree

benkozi:
@zmoon @bbakernoaa - Are there additional changes you'd like to see in this PR?

tests/test_ufs.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_deprecated_rrfs_cmaq_mm() -> None:
+    from monetio.models._rrfs_cmaq_mm import open_mfdataset  # noqa: F401
+
+
+def _compare_with_baseline_(actual: xr.Dataset, baseline_path: Path) -> None:
+    with xr.open_dataset(baseline_path) as baseline:
+        try:
+            assert actual.equals(baseline)
```
benkozi: Thinking about it, I should change this to an "is close" implementation. I'll do that in the near-term.

benkozi: I learned the difference between `equals` and `identical` (which check attributes). Interestingly, there seems to be a built-in tolerance with `identical` - nudging a variable by less than `1e-4` did not trigger a failure. Anyway, switched to `identical`. The tolerance seems acceptable to me.

bbakernoaa:
@benkozi This is passing the tests.  There is currently an issue with the AERONET tests that they are not passing.  I'm going to go ahead and approve this and I'll have a separate PR for fixing the CI tests for AERONET.

PR #247, opened 2025-07-22
Initial reader for UFS-Chem v1 files

mbruckner-work:
Adds an initial version of a specific reader for the UFS-Chem version 1 files. 
Reads gfs.t00z.atmf???.nc files and additional variables from the gfs.t00z.sfcf???.nc files if provided.


bbakernoaa:
@mbruckner-work do we need a separate reader for this? can we not use the same rrfs reader already? All ufs ones should be able to share.  the only difference would be if there is chemical mechanism calculations that would need to be done.

bbakernoaa:
Also I want to end the practice of creating readers with the `_` prefix.  We should be adding the readers to be exposed not hidden.

rschwant:
Closing this PR because I merged these updates into the ufs.py reader instead as requested by Barry and these updates are now in this PR: https://github.com/noaa-oar-arl/monetio/pull/261

PR #251, opened 2025-11-19
New TROPOMI reader.

blychs:
New TROPOMI reader, designed to work with any TROPOMI product.

This reader works by doing a recursive search over the variables. **It assumes that variable names are not repeated in different groups**.  The search is done every time a file is read, which is definitely an area that could be improved, but it does work pretty fast.

With minimal tweaks to MELODIES-MONET (I don't think anything needs to change in MONETIO) it could supersede the current _tropomi_l2_no2_mm.py reader.

Cheers
Pablo


noribeth-m:
@blychs In this reader, in the first function _open_one_dataset(), you have ds=ds.assign_coords() twice. Lines 38-44 and line 56.

blychs:
I believe it should be OK now

blychs:
@bbakernoaa , are we moving into using DataTrees? Will the requirements for MONETIO change then? Because if so, using netCDF4 wouldn't be necessary anymore.

bbakernoaa:
I'm still not sure that we need to do that either.  if we know the groups beforehand then we can supply the `group=` kwarg in the xarray open_dataset call.

monetio/sat/tropomi_l2.py
```diff
@@ -0,0 +1,403 @@
...
+    for group in groups:
+        try:
+            return _walktree_search(variable, netcdf_dataset[group], f"{path}/{group}")
+        except ValueError:
+            continue
+    raise ValueError(f"{variable} not found in {netcdf_dataset}")
+
+
+def _add_variable(variable, netcdf_dataset):
+    """Creates xr.DataArray formated for MELODIES-MONET.
```
bbakernoaa: spellcheck found a few.  please address them

bbakernoaa:
@blychs I'm going to approve but please address the code spell issue before I commit.  Let's take care of moving from netCDF4 in a separate PR.

blychs:
Hi Barry. I did. However, new issues were found when reading TROPOMI CO (mostly, the structure of the pressure in the CO product differs a lot from NO2 and HCHO, and the datatype of TROPOMI CO's quality_flag needs to be checked from the actual data instead of the netCDF4 file specification). So please give it a look before merging. I did the fixes as separate commits to avoid issues with that.
Cheers
Pablo

bbakernoaa (approved):

PR #261, opened 2026-03-06
Develop ufsreader

rschwant:
This PR just makes a small update to the WRF-Chem reader so that we can properly do the meterological evaluation and pull in the met calculations using WRF-python.

And this updates the UFS reader such that it can be used for UFS-Chem and UFS-AQM. So then we don't need a separate reader for UFS-Chem anymore. It also allows you to read in the surface data where a lot of the met calculations are stored (e.g., T2) and combine it together, so that you can evaluate this against the meteorological observations and also against chemistry observations.

Let me know if you see any problems.


monetio/models/ufs.py
```diff
@@ -46,11 +49,16 @@ def open_mfdataset(
     surf_only: boolean
         Whether to save only surface data to save on memory and computational
         cost (True) or not (False).
+    fname_sfc : string or list
+        Path to the sfc file in UFS. This file contains additonal variables
```
zmoon:
```suggestion
        Path to the sfc file in UFS. This file contains additional variables 
```

zmoon (commented):
You could also address the linter complaints (`pre-commit run --all-files`) and merge develop into your branch.

In the tests, most are unrelated (ICAP-MME website, https://usgodae.org/ftp/outgoing/nrl/ICAP-MME/, flaking), but `tests/test_ufs.py::test_open_mfdataset[surf_only=False]` failure could be related to your changes.

rschwant:
I figured out why the test_ufs.py is failing. The longitude values are different. Maggie due to the satellite evaluation wanted me to add this line to the reader for UFS-Chem:

    # modify longitude from 0-360 to -180-180 if needed
    dset["longitude"] = xr.where(
        dset["longitude"] >= 180, dset["longitude"] - 360, dset["longitude"]
    )

But this makes the test fail (Difference in variable: longitude) because the test example data is in 0 to 360 format and with my update the longitude data is in -180 to 180 format. The surface and aircraft data run the following code during the pairing which converts the data if needed to -180 to 180 format: _dataset_to_monet before pairing with obs.

I'm not sure that all of the satellite evaluation code does this conversion in the pairing though. I think it is okay to do this conversion in the model reader to make all the model data is eventually output in -180 to 180 format to be consistent with the observations. 

What do you think?

I can talk to Maggie next week if this is absolutely needed as well if you all prefer this not to be in the model readers for MONETIO?

zmoon:
I think it's okay to have this. You can just update the test (after loading the file to compare against, do the same conversion to it).

rschwant:
@zmoon, great I updated this! I even remade the file with only the update to the longitude correction. And then tested that file against all my updates and it worked well. So this is the only thing failing this test. I think it's best to apply this correction in the reader for longitude and then to also confirm in the pairing that it happens to be sure this correction is made throughout the code. Let me know if you see any other needed updates and thanks for the suggestions for the tests and pre-commit commands those are very useful!

monetio/models/ufs.py
```diff
@@ -20,6 +21,8 @@ def open_mfdataset(
     var_list=None,
     fname_pm25=None,
     surf_only=False,
+    fname_sfc=None,
+    sfc_varlist=["aod550"],
```
zmoon:
```suggestion
    sfc_varlist=None,
```
and in the function body near the top use
```python
if sfc_varlist is None:
    sfc_varlist = ["aod550"]
```

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

rschwant: I was just providing an example, but for consistency this should probably just be None, so I committed your suggestion. And there is no need for the if statement actually that would be strange.

zmoon: I thinking having it default to AOD 550 (when the arg is its default `None`) would've been fine, but up to you.

rschwant: I think it's okay to leave it as none. Any other updates on this?

zmoon: Alternatively, perhaps updating the later part to:
```python
ds_sfc = xr.open_mfdataset(fname_sfc, **kwargs)
if sfc_varlist is not None:
    ds_sfc = ds_sfc[sfc_varlist]
```
Otherwise you'd get `KeyError: None` if you gave sfc file name but not var list.

rschwant: Yes, opps I missed that, I'll fix this tomorrow.

rschwant:
@zmoon, I went with your first suggestion:

```python
if sfc_varlist is None:
    sfc_varlist = ["aod550"]
```

Because I'm not sure that all variables from that surface file can and we would want to merge in. So best for users to specify what they want and default to aod if nothing is provided.

Thanks!

monetio/models/ufs.py
```diff
@@ -46,17 +49,25 @@ def open_mfdataset(
     surf_only: boolean
         Whether to save only surface data to save on memory and computational
         cost (True) or not (False).
+    fname_sfc : string or list
+        Path to the sfc file in UFS. This file contains additional variables
+        and diagnostics not included in the standard atm files.
+    sfc_varlist : list
+        List of variables from the sfc file to include in output.
```
zmoon, line 56:
```suggestion
        List of variables from the sfc file to include in output.
        Defaults to ``['aod550']`` if not provided.
```

rschwant: good idea!

zmoon (approved):
Looking good to me

PR #266, opened 2026-03-17
Update ICAP-MME links

zmoon:
It moved a month or so ago

### DIFF
--- a/monetio/models/icap_mme.py
+++ b/monetio/models/icap_mme.py
@@ -31,9 +31,9 @@ including members that use MODIS and other observational total AOD in assimilati
 
 The data is 6-hourly, each file including multiple time steps (e.g. 21 or 25).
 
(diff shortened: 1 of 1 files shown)


monetio/models/icap_mme.py
```diff
@@ -31,9 +31,9 @@
 
 The data is 6-hourly, each file including multiple time steps (e.g. 21 or 25).
 
-Online visualizer: https://usgodae.org/metools/ensemble/
```
zmoon, line 34: Seems it is gone. @bbakernoaa do you know otherwise?

bbakernoaa: it might be.  I am not sure

zmoon: Ok well I'll just leave it off for now.

bbakernoaa (approved):

PR #274, opened 2026-04-20
ufs: Get lat/lon from grid_xt/yt if missing

zmoon:
See description of the situation in https://github.com/NCAR/MELODIES-MONET/issues/428#issuecomment-4284500076

xref: https://github.com/NCAR/MELODIES-MONET/issues/428

cc: @ytangnoaa

### DIFF
--- a/monetio/models/ufs.py
+++ b/monetio/models/ufs.py
@@ -231,7 +231,9 @@ def open_mfdataset(
         from ..util import _try_merge_exact
 
         # Add the processed pm2.5 species.
-        dset_pm25 = xr.open_mfdataset(fname_pm25, concat_dim="time", combine="nested", **kwargs)
+        dset_pm25 = _latlon2d(
+            xr.open_mfdataset(fname_pm25, concat_dim="time", combine="nested", **kwargs)
+        )
         dset_pm25 = dset_pm25.drop(
             labels=["lat", "lon", "pfull"]
         )  # Drop duplicate variables so can merge.
@@ -360,7 +362,7 @@ def open_mfdataset(
 
     # Read in additional variables from the sfc file
     if fname_sfc is not None:
-        ds_sfc = xr.open_mfdataset(fname_sfc, **kwargs)[sfc_varlist]
+        ds_sfc = _latlon2d(xr.open_mfdataset(fname_sfc, **kwargs))[sfc_varlist]
         ds_sfc = ds_sfc.rename({"grid_yt": "y", "grid_xt": "x"})
         if surf_only:  # Only expand into the zth dimension when surf_only is True,
             # so that the surface data are combined appropriately.
--- a/monetio/models/ufs.py
+++ b/monetio/models/ufs.py
@@ -14,11 +14,19 @@ def can_do(index):
         return False
 
 
-def _latlon2d(ds):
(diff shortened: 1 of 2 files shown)


tests/test_ufs.py
```diff
@@ -47,6 +47,68 @@ def test_open_mfdataset(data_dir: Path, test_data: DataForTest) -> None:
...
+    assert actual.sizes == orig.sizes
+
+    assert actual["latitude"].equals(orig["latitude"])
+    assert actual["longitude"].equals(orig["longitude"])
+
+
+def test_open_mfdataset_grid_xt_yt_dim_coords_only(data_dir: Path, tmp_path: Path) -> None:
+    """Test that open_mfdataset works when lat/lon are named grid_xt/grid_yt (1-D dim coords)."""
+    ufs_data_dir = data_dir / "ufs"
+    patt = "aqm.t12z.dyn.f*.nc"
```
benkozi: Minor - could be `path` :)

monetio/models/ufs.py
```diff
@@ -14,6 +14,53 @@ def can_do(index):
         return False
 
 
+def _latlon2d(ds):
```
benkozi: Minor - add types. Also could rename to something explicit like `_verify_or_update_spatial_coords`

zmoon:
CI failing due to current NCEI website issues and xdist collection failures.

monetio/models/ufs.py
```diff
@@ -14,6 +14,53 @@ def can_do(index):
...
+    """Sometimes lat/lon are missing (named grid_yt/grid_xt).
+    In this case we want to rename those vars.
+    If grid_yt/grid_xt vars are missing too (only dims present),
+    we raise an error.
+    """
+    if {"lat", "lon"} <= set(ds.variables):
+        pass
```
benkozi, line 32: Minor - could early return `ds` here instead of the big if block. This approach obviously works though

benkozi (approved):
Some minor comments! Otherwise looks good 👍

tests/test_ufs.py
zmoon: Added module-level variable

monetio/models/ufs.py
zmoon: Changed to `_ensure_latlon2d` and added docstring type info.

monetio/models/ufs.py
zmoon: With `if`-`elif` I think the single return makes more sense. Could separate the first `if` and early return `ds`, but I think the logic is a bit easier to follow with the single chain.

rschwant (approved):
This looks good and definitely adds a lot of flexibility for all conditions, so it's good to have this update.

I'm not sure why UFS output would move to using grid_xt/grid_yt as a dimension as well as a variable and they mean 2 different things. This is not typically done in models. So hopefully this is not their new format and just unique to what Youhua is working on?

benkozi:
> So hopefully this is not their new format and just unique to what Youhua is working on?

@rschwant - This is the "FV3 native grid" history file convention which does not follow CF, etc. and is disabled by default. I agree that it's confusing :)

@zmoon - The function name could be changed to `_handle_fv3_native_grid` if you wanted to be explicit

PR #277, opened 2026-05-08
Update model module names and drop a few older ones that have been superseded

rschwant:
Here I have deleted the older outdated _rrfs_cmaq_mm.py, camx.py, and cmaq.py files. I looked through these files and it looks like the corresponding updated files for ufs.py, _camx_mm.py, and _cmaq_mm.py are newer versions of these files and as such I think we should just delete these old versions, so there is less confusion, but let me know if you all disagree.

As I was comparing the diffs on these readers, I realized that the cmaq.py reader got updated here: https://github.com/noaa-oar-arl/monetio/commit/b116b0addd0838adf9b52f1b3162d68d7525b6ef for summing NOx as NO + NO2 instead of NOx as NO + NOx, but not the _cmaq_mm.py reader. So I added this in as well. So this is all consistent now.

Then I renmamed all the _model_mm.py files to model.py instead and updated the __init__.py file and tests with these changes. 

This connects with this PR in MELODIES MONET. https://github.com/NCAR/MELODIES-MONET/pull/430
We need to merge the PR in MELODIES MONET first before merging this in. But I'm submitting PRs together, so that you all can review and happy to make changes to either one as needed.


rschwant:
Just updated to fix the pre-commit test, I forgot to re-run this after fixing the tests.

zmoon (commented):
Renaming "private" readers to become public is one thing. For `cmaq` and `camx` though, which were public, the replacements could cause some breaking changes, so maybe we should release this as 0.4.0 instead of 0.3.3.

rschwant:
Yes, I can update in MELODIES MONET versions on the docs that we will move to 0.4 for this. That way it is clearer. Barry wanted us to move away from the private readers.

PR #278, opened 2026-05-17
Pandora text file read support

zmoon:
builds on #236 by @blychs 

adding:
* `read_txt` func that returns a dataframe
* profile support in the open-dataset funcs
  - stacking columns into a _z_ dim when extra layers besides layer 1 are detected
  - off by default, since it makes the file reading slower (use `layers=True` to opt in)
  - note the GEOMS format files seem to only offer total-column products, not partial column
* web API functions for discovering available data
* other tweaks


monetio/profile/pandora.py
```diff
@@ -0,0 +1,643 @@
...
+import warnings
+from glob import glob
+
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+_BASE_URL = "https://api.pandonia-global-network.org/v1"
+
+
```
zmoon:
```suggestion
```

monetio/profile/pandora.py
```diff
@@ -0,0 +1,641 @@
...
+                        "pan_id": pan_id,
+                        "spectrometer": spectrometer,
+                        "level": level,
+                    }
+                )
+
+    return pd.DataFrame(rows)
+
+
+def download(dates, *, location=None, prod="rfuh5"):
```
zmoon:
```suggestion
def download(dates, *, location=None, prod="fuh5"):
```
[probably](https://www.pandonia-global-network.org/assets/manuals/PGN_DataProducts_Readme_v1-8-10.pdf) should leave off the `r` as it's not considered part of the product code.

zmoon: maybe should leave as-is though, since the web API [does it this way too](https://api.pandonia-global-network.org/v1/tutorial/) and in fact it says
> ... the official codes: ['rout2', 'rwvt1', 'rfuh5', 'rnvh3', 'rnvs3', 'rsus1', 'rfus5', 'ffus5', 'fsus1', 'fout2', 'fwvt1', 'fnvs4', 'fnvs5', 'fnvh3', 'ffuh5', 'smca1']

PR #280, opened 2026-05-19
Generic model reader

zmoon:
intended for MM usage

```yaml
model:
  awesome-model:  # model label
    files: input/awesome-model-*.nc
    mod_type: generic
    is_global: True
    mod_kwargs:
      # Identify the dims, so we can rename
      x_dim: "X"
      y_dim: "Y"
      z_dim: "Z"
      time_dim: "T"
      # Identify key coords, so we can rename (and tweak if necessary)
      lat_var: "XLAT"
      lon_var: "XLON"
      time_var: "XTIME"  # needed if different from `time_dim`
      # Other options      
      surf_only: True  # select the level closest to surface
      surf_lev: -1  # the level closest to surface is the last one in this model

```

cc: @rschwant


bbakernoaa:
@zmoon shouldn't we just use v1? this is already solved

rschwant:
@zmoon, Zach are you intending on getting this in before MM v1.1, so that we can use this right away. If so, I think having this capability now is useful. It will take time to move to MONETIO v1 fully because we have to test that we have all the previous functionality fully updated in v1 and there is a lot of code to review.

zmoon:
> @zmoon shouldn't we just use v1? this is already solved

I don't think v1 has this (but feel free to point it out), though it would be easy to do there.

bbakernoaa:
v1 doesn't have a configuration like this per se but a strict format and capabilities for regridding etc are all required to follow either the CF or UGRID convention.  So the need to mark these attributes aren't really needed. And these are all accessed through a common `monetio.load("awesome_model", *kwargs)`.  Thats why I'm not sure that a yaml file for this.  

Also v1 does have the base model similar to your generic as well as a grid based class and point based class.  There are also "drivers" for XarrayDriver, PandasDriver, NCEPGRIBDriver, VirtualizarrDriver etc

zmoon:
> required to follow either the CF or UGRID convention. So the need to mark these attributes aren't really needed

The idea with this is to help you load data that *doesn't* follow conventions without having to preprocess it.

> Also v1 does have the base model similar to your generic as well as a grid based class and point based class. There are also "drivers" for XarrayDriver, PandasDriver, NCEPGRIBDriver, VirtualizarrDriver etc

Yes I do agree it would be easy to do this in the v1 framework, and better. But this could be nice to have for now.

bbakernoaa:
I also don't see a problem with having a requirement to say that we need data to follow a specific format. For instance if it follows a convention like COARDS/CF/MPAS/GRIB/UGRID/WRF etc MONETv3 handles all of these natively and a simply xr.open_dataset/xr.open_mfdataset would work. The generic yaml I just feel is going to be messy and require another user driven configuration to the already messy MM yaml configuration. 

Sooner or later we need to put a line in the sand

PR #281, opened 2026-05-26
Modernize documentation, enhance readers, and optimize cloud access

bbakernoaa:
This pull request introduces several important enhancements and documentation improvements to MONETIO, focusing on expanding data source support, improving developer and user documentation, and updating CI/CD workflows. The main changes include the addition of new readers for models, observations, satellites, and profile data; comprehensive documentation for these new sources and developer guidelines; and the introduction of a release workflow for automated publishing.

**Key changes:**

### 1. Expanded Data Source Support

* Added new model readers (`ncep_grib`, `gfs`, `gefs`, `gdas`, `rrfs`, `grib2`, `icap_mme`, `pardump`, `nam`, `rap`, `hrrr`) to the unified reader registry in `monetio/__init__.py`.
* Added new observation readers (`openaq_v2`, `openaq_v3`, `openaq_aws`, `madis`, `pams`, `ndbc`, `surfrad`, `solrad`, `tccon`, `ameriflux`) and profile readers (`amdar`, `iagos`, `umbc_aerosol`) to the registry. [[1]](diffhunk://#diff-0d528ac97907ab24f3fc458b403d81a32597c95a2f76339688e73c31120c2162R38) [[2]](diffhunk://#diff-0d528ac97907ab24f3fc458b403d81a32597c95a2f76339688e73c31120c2162R59)
* Added new documentation pages for satellites (`docs/satellites.md`) [[1]](diffhunk://#diff-28b4e8ef4eeeafee1d9132dd39282de9fc2c5898bc8383af15eb6d6260b2ebb2R1-R99) and profiles (`docs/profiles.md`) [[2]](diffhunk://#diff-662c46e284b76f4502bf5d26b4d6546c41ddcf2cf6abc66406ac7f95a3ee5af5R1-R76) and updated the navigation in `mkdocs.yml` to include them. [[3]](diffhunk://#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R28-R29) [[4]](diffhunk://#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2R39)

### 2. Documentation Improvements

* Overhauled the API reference (`docs/api.md`) to provide a clearer, more structured overview of core functions, utility methods, and base classes.
* Enhanced architecture documentation to clarify virtualization support (Kerchunk, Icechunk) and user interaction modes.
* Added a comprehensive developer's guide covering architecture, coding standards, testing, and contribution workflow (`docs/contributing/developers_guide.md`).
* Documented new and legacy observation/model sources, including usage examples for OpenAQ v3 and NESDIS VIIRS JRR. [[1]](diffhunk://#diff-8686497f7e75c1678bd0892894da48f22f1f4e412e9c0ca295901d1f6e468db2L55-R66) [[2]](diffhunk://#diff-8686497f7e75c1678bd0892894da48f22f1f4e412e9c0ca295901d1f6e468db2R97-R133) [[3]](diffhunk://#diff-28b4e8ef4eeeafee1d9132dd39282de9fc2c5898bc8383af15eb6d6260b2ebb2R1-R99) [[4]](diffhunk://#diff-4a20043f76a669ac926c5fe98226490f44635620f8f20e27f2703b7eb46a4b2eR46-R68)

### 3. CI/CD and Workflow Updates

* Introduced a new release workflow (`.github/workflows/release.yml`) for semantic versioning and automated PyPI publishing.
* Updated GitHub Actions to use newer versions of actions (`actions/checkout`, `actions/setup-python`) in CI, lint, and validation workflows for improved reliability and security. [[1]](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL40-R40) [[2]](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL90-R90) [[3]](diffhunk://#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2L21-R22) [[4]](diffhunk://#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2L34-R34) [[5]](diffhunk://#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2L43-R43)
* Added new entries to `.codespell-exclude` for domain-specific terms.

### 4. Virtualization and CLI Documentation

* Clarified the virtualization API and command-line interface, including new documentation for the `virtualizarr` command and its relation to `monetio.virtualize`. [[1]](diffhunk://#diff-27c8c22b45605a0ddf42ce8017b6b2a68cb3b10534c18bf4a781e230d7b92361R105-R106) [[2]](diffhunk://#diff-140eef3ba41bdcf401d507408084181f2c0ac627532b61e0f7906ea7cc926782L17-R29)

---

These changes collectively improve MONETIO's extensibility, usability, and maintainability for both users and contributors.


.github/workflows/ci.yml
DWesl: Why is this updated to v5, where the other update in this file is to v6?

```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/lint.yml
DWesl:
```suggestion
      - uses: actions/checkout@v6
```

DWesl, line 34:
```suggestion
      - uses: actions/checkout@v6
```

DWesl, line 43:
```suggestion
      - uses: actions/checkout@v6
```

.github/workflows/release.yml
DWesl:
```suggestion
      - uses: actions/checkout@v6
```

docs/contributing/developers_guide.md
DWesl, line 1: GitHub is a bit more effective at nudging people to this file if it's called [`docs/CONTRIBUTING.md`](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) instead of `docs/contributing/developers_guide.md`.

DWesl, line 47: Might be helpful to provide a link to the description of NumPy-style docstrings
```suggestion
We use [NumPy-style docstrings](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard). Every public function and class should be documented.
```

docs/api.md
DWesl, line 9: The triple-colon isn't a Markdown construct I recognize: what is this supposed to do?

DWesl, line 3: I'm guessing that page is generated by `docs/gen_ref_pages.py`

monetio/readers/actris.py
```diff
@@ -210,6 +210,14 @@ class ACTRISReader(PointReader):
     def open_dataset(
         self,
         files: str | list[str] | None = None,
+        use_virtualizarr: bool = False,
```
DWesl, line 213: This will change the API, though I can see the reasoning for this line being early, and it should at least fail loudly if people don't notice

monetio/readers/ameriflux.py
```diff
@@ -0,0 +1,115 @@
...
+                df["time"] = pd.to_datetime(ts_str, format="%Y")
+            elif len(ts_str.iloc[0]) == 6:
+                df["time"] = pd.to_datetime(ts_str, format="%Y%m")
+            else:
+                df["time"] = pd.to_datetime(ts_str)
+
+        # 2. Rename coordinates if present (often in a separate metadata file,
+        # but sometimes included or can be passed)
+        # Note: BASE files themselves often DON'T have Lat/Lon in every row.
+        # They are usually in the BADM (metadata) files.
```
DWesl, line 111: It looks like the [CF Discrete Sampling Geometries conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#discrete-sampling-geometries) might be useful for handling the location metadata in XArray, though that doesn't translate well to Pandas

monetio/readers/chimere.py
```diff
@@ -42,10 +45,16 @@ def open_dataset(
             Whether to use VirtualiZarr to create a virtual Zarr dataset, by default False.
         virtualizarr_file : str or None, optional
             Path to save/load the VirtualiZarr reference JSON file, by default None.
+        virtualizarr_parser : str or None, optional
+            The VirtualiZarr parser to use (e.g. 'hdf5', 'netcdf3', 'zarr', 'grib2').
         virtualizarr_backend : str, optional
```
DWesl, line 50: You can list the options as part of the [parameter type declaration](https://numpydoc.readthedocs.io/en/latest/format.html#parameters):

```suggestion
        virtualizarr_backend : {"kerchunk", "icechunk"}
```

monetio/readers/jpss_cris.py
```diff
@@ -0,0 +1,129 @@
...
+        y_dim=["nscan", "rows", "y"],
+        x_dim=["nlon", "cols", "x"],
+        z_dim=["npress", "lev", "level", "z"],
+    )
+
+    # 2. Variable renaming
+    mapping = {
+        "temperature": "temperature",
+        "water_vapor": "specific_humidity",
+        "pressure": "pressure",
```
DWesl, line 113: What is your reasoning for including the renaming of "temperature" to "temperature" and "pressure" to "pressure"?

DWesl (commented):
Looks solid, but I don't know the code well enough to say more than that

.github/instructions/bash.instructions.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+## 1. Google Shell Style Guide Foundations
+* **Interpreter Shebang:** Always start scripts with `#!/bin/bash`.
+* **The Main Function:** Wrap all execution logic in a `main` function. The final line of the script must explicitly invoke `main "$@"`.
+* **Function Declarations:** Declare functions using `my_func() { ... }`. Never use the legacy `function` keyword.
+* **Local Variables:** Always declare variables inside functions using `local` (e.g., `local var_name="value"`). Keep variables local to restrict scope clutter.
+* **Variable Expansion:** Always quote variable expansions to prevent word splitting and globbing issues. Prefer `"${var}"` over `$var`.
+* **Tests and Evaluations:** Strictly use `[[ ... ]]` for evaluations and condition checks instead of legacy `[` or `test`.
+* **Command Substitution:** Always use `$(command)` for command substitution. Never use legacy backticks (\`command\`).
```
DWesl: Triple-ticks separated by spaces from single-ticks gives: ``` `command` ```.  

Not sure what Copilot expects
```suggestion
* **Command Substitution:** Always use `$(command)` for command substitution. Never use legacy backticks (``` `command` ```).
```

.github/instructions/bash.instructions.md
```diff
@@ -0,0 +1,53 @@
...
+Scripts must match NCO's multi-tiered environment pipeline. Variables must be exported down the execution chain, never passed as positional command-line arguments.
+
+* **J-Jobs (`jobs/` directory):**
+  * Naming: Must adhere to `JAAAAA` (Capital `J` followed by all caps, no file extension).
+  * Purpose: Sets operational paths (`$DATA`, `$COMIN`, `$COMOUT`, `$EXECmodel`, `$USHmodel`) and temporal limits (`$PDY`, `$cycle`), clears/initializes the work area, and invokes the matching ex-script.
+  * Constraint: Never alter or overwrite variables established at the J-Job level further downstream.
+* **Ex-Scripts (`scripts/` directory):** The core driver logic for a specific application task. Called cleanly by the J-Job wrapper.
+* **Ush-Scripts (`ush/` directory):** Utility scripts called by ex-scripts to execute modular, repetitive workflows.
+
+---
```
DWesl, line 33: I don't know that this section is relevant to this repository.  Is this designed to be standard across NOAA repos?

.github/instructions/cpp.instructions.md
```diff
@@ -0,0 +1,54 @@
+---
+description: NWS Office of Modeling and Development - C++ Guidelines (Forge Persona)
+applyTo: "**/*.{cpp,hpp,c,h,cc,hh,cxx,hxx}"
```
DWesl, line 3: Is this case-sensitive enough to need `**/*.[CH]` specified separately?

.github/instructions/cpp.instructions.md
```diff
@@ -0,0 +1,54 @@
...
+
+You operate as **Forge**, a Principal High-Performance Systems Architect specializing in Modern C++ (C++23) and the Kokkos performance portability ecosystem. Your mission is to architect high-performance, memory-safe meteorological systems that interoperate with legacy Fortran kernels while strictly adhering to NCO HPC Environment Equivalence (EE2) standards.
+
+---
+
+## 1. Core Principles & Memory Safety
+* **Modern Standard (C++23):** Default to C++23 features. Code must leverage zero-cost abstractions, `constexpr`/`consteval` evaluations, and C++20/23 Concepts to maximize throughput.
+* **Strict RAII:** Absolute zero tolerance for raw owning pointers (`new`/`delete`) or manual memory management.
+* **Const Correctness:** Everything that can be `const` MUST be `const`.
+* **Error Handling (EE2):** Fail fast and gracefully. For discoverable errors (like missing operational input files), use `std::expected` or exceptions to catch and print descriptive `FATAL ERROR:` or `WARNING:` messages before a segmentation fault can occur.
```
DWesl, line 16: Are there guidelines about whether messages should go to `std::cout`, `std::cerr`, `std::clog`, `STDOUT`, or `STDERR` (using `iostream` or `cstdio` functions)?

.github/instructions/fortran.instructions.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+# Fortran Guidelines: Flux Protocol & EE2 Compliance
+
+You operate as **Flux**, a Principal High-Performance Scientific Computing Architect specialized in Modern Fortran (2018+). Your core mission is to architect and optimize numerical weather prediction kernels that achieve maximum performance, portability, and hybrid parallelism while adhering strictly to NCO HPC Environment Equivalence (EE2) standards.
+
+---
+
+## 1. Core Principles & Memory Safety
+* **Modern Standards:** Default to Modern Fortran (2018+ standards). Strictly eliminate and refactor legacy Fortran constructs (e.g., `GOTO` statements, `COMMON` blocks, implicit typing).
```
DWesl, line 13: I'm guessing the "e.g." is so you don't have to list out arithmetic/three-way if and shared `DO` labels?

.github/instructions/fortran.instructions.md
```diff
@@ -0,0 +1,59 @@
...
+* **Strict Typing:** `IMPLICIT NONE` is mandatory in every module and subroutine without exception.
+* **Explicit Intent:** Every single dummy argument must have an explicitly declared `INTENT` (e.g., `INTENT(IN)`, `INTENT(OUT)`, `INTENT(INOUT)`).
+* **Precision Portability:** NEVER use `REAL*8`, `REAL*4`, or generic `DOUBLE PRECISION`. Define a central type-kind module (e.g., `mod_kinds`) leveraging `iso_fortran_env` (such as `wp = real64`) and import it (`use mod_kinds, only: wp`) across the entire repository to remain precision-agnostic.
```
DWesl, line 16: Does this also imply the avoidance of algorithms producing a specific number of digits of precision, or defining relevant constants in `mod_kinds` and having the algorithms use those to achieve the desired precision?

.github/instructions/fortran.instructions.md
```diff
@@ -0,0 +1,59 @@
...
+* **Pure Procedures:** Default to writing `PURE FUNCTION` or `ELEMENTAL SUBROUTINE` components to ensure complete thread safety, side-effect-free execution, and aggressive compiler optimizations.
+* **Vectorization:** Prefer whole-array syntax operations (e.g., `A = B + C`) over manual, nested `DO` loops to enable automatic compiler SIMD vectorization, unless a loop requires precise OpenMP/OpenACC tuning.
+
+---
+
+## 3. Operational Integrity & I/O Boundaries
+* **Separation of Concerns (Compute-Only):** Core numerical and compute kernels (placed in `src/`) must NEVER write directly to `stdout`, standard print streams, or local files. All direct operational I/O, status logging, or data output must be isolated strictly to the driver application level (placed in `app/`).
+* **HPC Data Formats:** For intensive model arrays and restarts, utilize parallel-aware structures such as NetCDF or HDF5. Light metadata or static tables may use structured CSV or clean binary layout.
```
DWesl, line 33: Should something similar be in the C++ section?  (Perhaps with the addition of Zarr?)

.github/instructions/python.instructions.md
```diff
@@ -0,0 +1,74 @@
...
+* **Deep Learning & AI:** Use **`pytorch`** (and the `torch` ecosystem) or **`jax`** for neural network development and execution.
+
+---
+
+## 3. Data Engineering Protocols (Aero Core)
+When writing standard meteorological data pipelines, enforce these four competing goals: Flexibility (Eager/Lazy execution), Maintainability, Provenance, and Two-Track Visualization.
+
+### A. The "Optional Dask" Architecture
+* **Backend Agnostic:** Functions must accept generic `xr.DataArray` or `xr.Dataset` inputs. Never assume or force data to be exclusively Dask-backed or NumPy-backed.
+* **No Hidden Computes:** NEVER call `.compute()`, `.load()`, or `.values` inside a downstream processing function. This immediately breaks laziness for Dask users and risks memory crashes on the login or compute nodes.
```
DWesl:
```suggestion
* **No Hidden Computes:** NEVER call `.compute()`, `.load()`, `.persist()`, or `.values` inside a downstream processing function. This immediately breaks laziness for Dask users and risks memory crashes on the login or compute nodes.
```

.github/instructions/python.instructions.md
```diff
@@ -0,0 +1,74 @@
...
+* **No Forced Chunking:** Do not hardcode explicit `.chunk()` calls inside calculation functions. Chunking operations belong strictly at the initial I/O stage or must be passed as an optional parameter.
+* **High-Performance Vectorization:** Native Python loops are strictly forbidden for numerical operations. Use `xarray.apply_ufunc` with `dask='parallelized'` capability to natively support both eager and lazy backends simultaneously.
+
+### B. Scientific Hygiene & Documentation
+* **Documentation Standard:** By default, every function must feature a strict **NumPy-style docstring** detailing `Parameters`, `Returns`, and `Examples`.
+* **Type Hinting:** Use explicit modern type hints (e.g., `xarray.DataArray`, `xarray.Dataset`). Do not use specific internal backend types like `dask.array`.
+* **Provenance Tracking:** Always update metadata attributes (e.g., `ds.attrs['history']`) when modifying or transforming a dataset. Never drop coordinate variables or critical spatial dimensions during processing.
```
DWesl, line 44: Should this also mention the CF `cell_methods` variable attribute or the ACDD `date_modified`/`date_metadata_modified`?  Should it also track CF `ancillary_variables`?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,125 @@
...
+* **Constraint:** You must halt and acquire explicit user approval after generating each document before proceeding to the next. Do not write implementation code until the `tasks.md` file is finalized.
+
+### 2.2 Authoring Format Standard (Human + Machine Readable)
+All instruction files should be written so they are easy for humans to scan and easy for agents to parse.
+
+* Use short sections with stable headings and explicit scope statements.
+* Use one rule per bullet with a bold keyword prefix (for example, `**Error Handling:**`).
+* Avoid malformed markdown, dangling code fences, and mixed inline list markers.
+* Keep examples minimal, executable, and clearly fenced.
+* Prefer imperative language (`must`, `must not`, `never`, `always`) for enforceable rules.
```
DWesl, line 52: Is this a request for a specific subset of [RFC 2119](https://www.ietf.org/rfc/rfc2119.html)?  Referencing previous efforts on the subject can simplify understanding

.github/copilot-instructions.md
```diff
@@ -0,0 +1,125 @@
...
+* Keep references to external standards as markdown links near the rule that depends on them.
+
+---
+## 3. Strict EE2 Environment & Safety Constraints
+* **Path Sanitation:** Do not append or prepend the current directory (`.`), `$USHmodel`, or `$EXECmodel` to the system `$PATH`. Invoke scripts and executables explicitly via their fully qualified environment pathing.
```
DWesl: I'd prefer slightly different capitalization due to the bash spelling; the current label looks more suited to rules like "create neither directories nor files with spaces in the name", though the description here is clear.
```suggestion
* **PATH Sanitation:** Do not append or prepend the current directory (`.`), `$USHmodel`, or `$EXECmodel` to the system `$PATH`. Invoke scripts and executables explicitly via their fully qualified environment pathing.
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,125 @@
...
+* **No Background Processes:** Never use the ampersand `&` to run processes in the background. The scheduler loses control of backgrounded children, violating operational policy.
+* **Context-Aware Environments (Modules vs. Conda/Apptainer):**
+  * **If writing an Operational Script (J-Job, Ex-script):** Explicitly load the target machine's software stack using `module purge`, followed by `module use "${REPO_ROOT}/modelfiles"`, and `module load ${MACHINE}`.
+  * **If writing a Development/Bootstrap Script:** It is permissible to script the creation and activation of a Python `venv` or `.conda` environment (e.g., `conda env create -f environment.yml`), or to wrap execution inside an Apptainer container (`apptainer exec container.sif python script.py`).
+  * **Rule:** Never leak development `.conda` or `venv` activation commands into operational J-Jobs.
```
DWesl, line 62: Why has conda been appearing with a leading dot and venv without?  I've seen `.venv` as an acutal root directory name for virtual environments, but I'm not sure I've seen the corresponding directory for conda

.github/copilot-instructions.md
```diff
@@ -0,0 +1,125 @@
...
+* **Path Sanitization:** Prevent directory traversal attacks. Any user or downstream-supplied path input must be rigorously sanitized before being passed to shell commands or file I/O operations.
+* **Data Privacy:** Never log or print Personally Identifiable Information (PII) or sensitive infrastructure layouts to standard application logs.
+
+### 7. Version Control & CI/CD Pipelines
+When assisting with Git workflows, Pull Requests, or CI/CD configuration files (GitHub Actions, Jenkins), apply these rules:
+
+* **Conventional Commits:** When generating commit messages, use the Conventional Commits specification (e.g., feat:, fix:, refactor:, perf:).
+* **Atomic Changes:** Encourage atomic, single-purpose commits to keep the repository history bisectable.
+* **Test Generation First:** When writing new CI/CD workflow files, always ensure that testing and linting jobs are executed before any compilation or deployment steps. Assume a strict gateway where failing tests block operational merges.
```
DWesl, line 106: That is an interesting order.  I can see doing linting jobs first, but I'm not sure how to do testing before compilation.  I can see lint, simple compile, simple test, complex/matrix compile, complex/matrix test, deploy as an order, so that CI fails with as little compute time used as possible

monetio/readers/actris.py
DWesl: Would it be worth marking these as keyword-only?  It's an even bigger API break, but it'll definitely catch everyone who didn't notice

monetio/readers/actris.py
DWesl, line 257: At this point it's getting at least close to needing an ["Other Parameters"](https://numpydoc.readthedocs.io/en/latest/format.html#other-parameters) section.  I'm not sure where you want to put that line

monetio/readers/amdar.py
```diff
@@ -0,0 +1,143 @@
...
+            Path to the Icechunk repository, by default None.
+        use_dask : bool, optional
+            Whether to use Dask for lazy loading, by default False.
+        **kwargs : dict
+            Additional arguments passed to Xarray.
+
+        Returns
+        -------
+        xr.Dataset or pd.DataFrame
+            UGRID xarray dataset by default, or DataFrame if ``as_xarray=False``.
```
DWesl, line 71: UGRID is an interesting choice for aircraft data.  I think [CF Discrete Sampling Geometry "trajectory"](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#discrete-sampling-geometries) would be more natural, perhaps as given by the [NCEI (formerly NODC) netCDF templates](https://www.ncei.noaa.gov/data/oceans/ncei/formats/netcdf/v2.0/index.html) for trajectory data, or the [NCAR Research Aviation Facility conventions](https://field.eol.ucar.edu/docs/raf/Software/netCDF.html) would be more natural

monetio/readers/amdar.py
```diff
@@ -0,0 +1,143 @@
...
+            Harmonized dataset.
+        """
+        # Mapping for aircraft data
+        mapping = {
+            "latitude": "latitude",
+            "longitude": "longitude",
+            "observationTime": "time",
+            "altitude": "altitude",
+            "pressure": "pressure",
+            "temperature": "temperature",
```
DWesl, line 155: The renaming logic below specifically skips where the key is not in the dataset or the value is in the dataset.  With both values the same, one will always be true.

monetio/readers/aqs.py
```diff
@@ -106,24 +105,302 @@
...
+        "WIND": "WIND_",
+        "TEMP": "TEMP_",
+        "RHDP": "RH_DP_",
+        "WS": "WIND_",
+        "WDIR": "WIND_",
+    }
+    code = mapping.get(p, p + "_")
+
+    url = f"{beginning}{code}{year}.zip"
+    fname = f"{fname_prefix}{code}{year}.zip"
```
DWesl: Given the definitions above, the following would appear to give the same answer with less repitition:
```suggestion
    fname = f"{fname_prefix}{code}{year}.zip"
    url = f"{AQS_BASE_URL}{fname}"
```

monetio/readers/base.py
```diff
@@ -67,6 +67,40 @@ def harmonize(self, ds):
...
+        if len(time_dims) == 1 and time_dims[0] in ds.dims:
+            try:
+                ds = ds.swap_dims({time_dims[0]: "time"})
+            except Exception:
+                pass
+            return ds
+
+    if "time" in ds.variables and "time" not in ds.coords:
+        ds = ds.set_coords("time")
+        return _ensure_time_dimension(ds)
```
DWesl, line 96: Is there some reason you cannot move this section before the `if "time" in ds.coords:` check?

monetio/readers/calipso.py
```diff
@@ -0,0 +1,142 @@
...
+        y_dim=["nray", "profile", "fakeDim0", "N_Profiles"],
+        z_dim=["nbin", "range_bin", "fakeDim1", "N_Range_Bins"],
+    )
+
+    # 2. Handle Time
+    if "Profile_Time" in ds.variables:
+        # Profile_Time is often seconds since a reference
+        # But for now, let's just make sure it's a coord
+        if "time" not in ds.coords:
+            if "Profile_Time" in ds.variables:
```
DWesl: You've already got this condition four lines up
```suggestion
```

monetio/readers/drivers.py
```diff
@@ -142,12 +257,27 @@ def expand_paths(path_input: str | list[str], fs=None) -> list[str]:
...
-                # fs.glob usually returns paths without the protocol (e.g. 'bucket/file.nc')
-                if path_input.startswith("s3://") and files and not files[0].startswith("s3://"):
-                    files = [f"s3://{f}" for f in files]
 
+                # Ensure paths have the protocol if the input had it
                 if not files:
                     raise FileNotFoundError(f"No files found matching pattern: {path_input}")
+
+                # fsspec.unstrip_protocol is the standard way to restore the protocol
+                # but it might not be available or consistent across all versions/fs.
```
DWesl, line 316: [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) from the standard library might also be able to get the protocol (there called "scheme") from `path_input`, and [`urllib.parse.urlunsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit) or [`urllib.parse.urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) (the behavior the docs warn about is the intended behavior here, so it should be fine)

monetio/readers/drivers.py
```diff
@@ -142,12 +257,27 @@ def expand_paths(path_input: str | list[str], fs=None) -> list[str]:
...
+                # Manual fix for common cases in monetio:
+                protocol = ""
+                if path_input.startswith("s3://"):
+                    protocol = "s3://"
+                elif path_input.startswith("http://"):
+                    protocol = "http://"
+                elif path_input.startswith("https://"):
+                    protocol = "https://"
```
DWesl: Slightly less code, and adds `ftp` to the list of protocols, paralleling `FileUtility.get_fs` above.
```suggestion
                possible_protocol, _host_path = path_input.split("://", 1)
                if possible_protocol in ("s3", "http", "https", "ftp"):
                    protocol = f"{possible_protocol}://"
```

monetio/readers/drivers.py
```diff
@@ -161,18 +291,136 @@ def expand_paths(path_input: str | list[str], fs=None) -> list[str]:
...
     The unified driver for opening gridded data (NetCDF, GRIB, HDF).
-    Supports S3 via fsspec.
+    Supports S3 via obstore or fsspec.
     """
 
+    def _open_grib2io_native(
+        self,
+        file_list: list[str],
+        xr_kwargs: dict,
+        preprocess: Callable | None,
```
DWesl: I'm reasonably certain of the signature of the `preprocess` function:
```suggestion
        preprocess: Callable[[xr.Dataset], xr.Dataset] | None,
```
Also, given the `| None` part of the signature, I'd expect this to have a default value of `None` and therefore be two or three lines later unless you plan to make either no or three retries default

monetio/readers/epa_utils.py
```diff
@@ -15,18 +16,20 @@
...
+def convert_statenames_to_abv(
+    df: Union[pd.DataFrame, "dd.DataFrame"],
+) -> Union[pd.DataFrame, "dd.DataFrame"]:
     """
     Convert full state names to two-letter abbreviations.
 
     Parameters
     ----------
-    df : pd.DataFrame
+    df : Union[pd.DataFrame, dd.DataFrame]
```
DWesl, line 27: [The numpydoc standard](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) tends to use "or" instead of the new Union construct.  From discussions in the numpydoc repository, this is as much due to concerns about the readability of docstrings by new users who know the `help` function but not much else as it is to numpydoc predating type annotations by several years.

At least one linter only does text comparisons of type annotations and docstring types, which is simpler with this style

monetio/readers/gems.py
```diff
@@ -0,0 +1,195 @@
...
+        group : str or list of str, optional
+            The NetCDF group(s) to open. If a list is provided, groups will be merged.
+            If None, common GEMS groups will be opened:
+            - "Data Fields"
+            - "Geolocation Fields"
+            - "Metadata"
+        variable_dict : dict, optional
+            Dictionary mapping variable names to processing options (scale, minimum, maximum).
+        **kwargs : dict
+            Additional arguments passed to XarrayDriver.open.
```
DWesl: This should probably be a link to `XarrayDriver.open`.  I think you're using a different documentation generator than I'm used to, so this is as much as I know.
```suggestion
            Additional arguments passed to `XarrayDriver.open`.
```

monetio/readers/amdar.py
```diff
@@ -131,7 +171,9 @@ def harmonize(self, ds: xr.Dataset) -> xr.Dataset:
         # Handle time if needed
         if "time" in ds.variables:
             if ds["time"].attrs.get("units") == "seconds since 1970-01-01 00:00:00.0 +0000":
-                ds["time"] = pd.to_datetime(ds["time"].values, unit="s")
+                # Convert to datetime64[ns] lazily by using the epoch offset
+                # 1970-01-01 is the default epoch for datetime64
+                ds["time"] = (ds["time"] * 1e9).astype("datetime64[ns]")
```
DWesl, line 176: I would usually do `ds["time"] = ds["time"].astype("datetime64[s]").astype("datetime64[ns]")`, but this works too.

monetio/readers/drivers.py
```diff
@@ -585,44 +578,164 @@ def open(
 
             # In MONETIO, GRIB virtual references must be routed through grib2io.
             if parser_name == "grib2":
-                xr_kwargs["engine"] = "grib2io"
+                try:
+                    import grib2io
+                    from grib2io.kerchunk import ReferenceGenerator
+                except ImportError:
+                    raise ImportError("grib2io is required for GRIB2 VirtualiZarr reading.")
```
DWesl, line 637: `grib2io[kerchunk]` might be clearer about what to install

monetio/readers/drivers.py
```diff
@@ -132,22 +132,72 @@ def _open_via_icechunk(vds, icechunk_url: str, virtualizarr_file: str | None) ->
...
+        container = icechunk.VirtualChunkContainer(
+            url_prefix=prefix,
+            store=store_conf,
+        )
+        config.set_virtual_chunk_container(container)
+
+    authorize = {prefix: None for prefix in unique_prefixes}
+
+    if icechunk_url.startswith("s3://"):
+        parts = icechunk_url[5:].split("/", 1)
```
DWesl, line 182: Would [`urllib.parse.urlsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) be a useful way to do this?  
```python
_, bucket, prefix, _, _ = urlsplit(icechunk_url)
```

monetio/readers/drivers.py
```diff
@@ -728,6 +786,14 @@ def open(
...
                 xr_kwargs.pop("icechunk_url", None)
                 xr_kwargs.pop("icechunk_repo", None)
+                xr_kwargs.pop("use_icechunk", None)
+                xr_kwargs.pop("store_path", None)
+                xr_kwargs.pop("max_scan_attempts", None)
+                xr_kwargs.pop("network_timeout", None)
+                xr_kwargs.pop("max_concurrent_requests", None)
+                xr_kwargs.pop("storage_options", None)
+                xr_kwargs.pop("filters", None)
+                xr_kwargs.pop("max_workers", None)
```
DWesl: Why are these in a separate block of individual `pop` calls rather than a loop over a list like two lines before and the previous function?

Would it be worth pulling the common elements of those lists into a global constant?

Actually, since you're not using the value, why not just delete the keys?
```python
for key in itertools.chain(mfdataset_keys, dataset_driver_keys, ...):
    try:
        del xr_kwargs[key]
    except KeyError:
        pass
```

monetio/readers/drivers.py
DWesl: Alternately:
```suggestion
                url_parts = urllib.parse.urlsplit(path_input)
                if url_parts.scheme in ("s3", "http", "https", "ftp"):
                    protocol = f"{url_parts.scheme:s}://"
```

monetio/readers/drivers.py
```diff
@@ -585,44 +578,164 @@ def open(
...
+                for key in mfdataset_keys:
+                    xr_kwargs.pop(key, None)
+
+                xr_kwargs.pop("engine", None)
+                xr_kwargs.pop("use_virtualizarr", None)
+                xr_kwargs.pop("virtualizarr_backend", None)
+                xr_kwargs.pop("virtualizarr_file", None)
+                xr_kwargs.pop("virtualizarr_parser", None)
+                xr_kwargs.pop("icechunk_url", None)
+                xr_kwargs.pop("icechunk_repo", None)
```
DWesl:
```suggestion
                virtualizarr_keys = [
                    "use_virtualizarr",
                    "virtualizarr_backend",
                    "virtualizarr_file",
                    "virtualizarr_parser",
                ]
                import itertools
                for key in itertools.chain(
                    mfdataset_keys, virtualizarr_keys, ["engine", "icechunk_url", "icechunk_repo"]
                ):
                    try:
                        del xr_kwargs[key]
                    except KeyError:
                        pass
```

monetio/readers/madis.py
```diff
@@ -136,7 +176,9 @@ def harmonize(self, ds: xr.Dataset) -> xr.Dataset:
         # Handle time if it's in seconds since epoch
         if "time" in ds.variables:
             if ds["time"].attrs.get("units") == "seconds since 1970-01-01 00:00:00.0 +0000":
-                ds["time"] = pd.to_datetime(ds["time"].values, unit="s")
+                # Convert to datetime64[ns] lazily by using the epoch offset
+                # 1970-01-01 is the default epoch for datetime64
+                ds["time"] = (ds["time"] * 1e9).astype("datetime64[ns]")
```
DWesl: Perhaps the following to avoid converting to float if unnecessary?
```suggestion
                ds["time"] = (ds["time"] * 1_000_000_000).astype("datetime64[ns]")
```
[Underscores in integer literals added in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep515)

tests/test_madis_modern.py
```diff
@@ -0,0 +1,49 @@
...
+    # Check that time was converted correctly
+    assert ds_eager.time.dtype == "datetime64[ns]"
+    assert ds_eager.time.values[0] == np.datetime64("2023-01-01T00:00:00")
+
+
+def test_madis_dataframe_lazy():
+    # Verify as_xarray=False with use_dask=True
+    # We can't easily use tmp_path with dask.dataframe for a single file without actual NetCDF
+    # but we can mock the reader logic if needed or just use the file created above.
+    pass
```
DWesl, line 49: Are you planning to actually fill in a function body at some point?  You added a dask task graph checker a few weeks ago that would likely work.

tests/test_madis_modern.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # 2. Eager Path (NumPy)
+    ds_eager = reader.open_dataset(str(fn), use_dask=False, as_xarray=True)
+    assert isinstance(ds_eager.temperature.data, np.ndarray)
+
+    # 3. Lazy Path (Dask)
+    ds_lazy = reader.open_dataset(str(fn), use_dask=True, as_xarray=True)
+    assert hasattr(ds_lazy.temperature.data, "dask")
+
+    # 4. Assert Identity
```
DWesl:
```suggestion
    # 4. Assert Identical
```

monetio/readers/goes.py
```diff
@@ -267,17 +269,28 @@ def _proj_inv(xv: np.ndarray, yv: np.ndarray, p_srs: str) -> tuple:
 
         Returns
         -------
-        tuple
+        tuple[np.ndarray, np.ndarray]
```
DWesl, line 272: I suppose that's more consistent with the type annotations than
```suggestion
        latitude, longitude : np.ndarray
```

Given the style you're applying, do you have a linter checking the types in the annotations against those in the docstring?

tests/test_cmaq.py
```diff
@@ -135,3 +135,28 @@ def test_cmaq_diagnostics():
...
+    ds_eager = _add_ioapi_latlon(ds.copy(), proj4)
+
+    # 3. Lazy execution
+    ds_lazy = ds.chunk({"x": 3, "y": 3})
+    ds_lazy_out = _add_ioapi_latlon(ds_lazy, proj4)
+
+    # 4. Assertions
+    from dask.array import Array
+
+    assert isinstance(ds_lazy_out.latitude.data, Array)
```
DWesl, line 159: Another style of laziness check.  At this point you should create either a `assert_array_lazy` or a `check_array_laziness` function in a testing utilities module.

tests/test_goes_opt.py
```diff
@@ -81,3 +81,49 @@ def test_goes_opt_values():
...
+            "semi_major_axis": 6378137.0,
+            "semi_minor_axis": 6356752.31414,
+            "inverse_flattening": 298.257222103,
+            "latitude_of_projection_origin": 0.0,
+            "longitude_of_projection_origin": -75.0,
+            "sweep_angle_axis": "x",
+            "grid_mapping_name": "geostationary",
+        },
+    )
+    ds["goes_imager_projection"] = proj
```
DWesl: Since you don't use `proj` later, you can assign directly to `ds["goes_imager_projection"]`:
```suggestion
    ds["goes_imager_projection"] = (
        (),
        np.int32(0),
        {
            "perspective_point_height": 35786023.0,
            "semi_major_axis": 6378137.0,
            "semi_minor_axis": 6356752.31414,
            "inverse_flattening": 298.257222103,
            "latitude_of_projection_origin": 0.0,
            "longitude_of_projection_origin": -75.0,
            "sweep_angle_axis": "x",
            "grid_mapping_name": "geostationary",
        },
    )
```

monetio/readers/drivers.py
```diff
@@ -132,22 +132,72 @@ def _open_via_icechunk(vds, icechunk_url: str, virtualizarr_file: str | None) ->
     xr.Dataset
         The dataset opened from the Icechunk store.
     """
+    # If the local directory already exists, clear it for a clean, idempotent run
+    if not icechunk_url.startswith("s3://") and os.path.exists(icechunk_url):
+        import shutil
+        try:
+            shutil.rmtree(icechunk_url)
+        except Exception:
+            pass
```
DWesl, line 141: Is this expecting subsequent loads to go through a Zarr interface?  Is that mentioned in the docs somewhere?

.github/instructions/bash.instructions.md
DWesl: Double-backticks around the command with backticks will produce the formatting (`` `command` ``), though it needs spaces between the double backticks and single backticks to tell them apart
```suggestion
* **Command Substitution:** Always use `$(command)` for command substitution. Never use legacy backticks (`` `command` ``).
```

DWesl (commented):
Very broad scope, but I think I looked at everything

.github/instructions/bash.instructions.md
bbakernoaa: We can remove the non domain languages from this as it should be Python only.

.github/instructions/bash.instructions.md
bbakernoaa: I would like to include this for anything that is potentially R2O as I hope this will be

.github/instructions/cpp.instructions.md
bbakernoaa: Can be removed as the bash one was.

.github/instructions/cpp.instructions.md
bbakernoaa: Same

.github/instructions/fortran.instructions.md
bbakernoaa: Same

.github/instructions/fortran.instructions.md
bbakernoaa: There are very few routines that specify a specific number of precision. I don't know if I've ever actually seen this in the wild.

.github/instructions/fortran.instructions.md
bbakernoaa: There should be something similar in c++ but it may be better more generally to cover other languages too.

.github/instructions/python.instructions.md
bbakernoaa: Not sure what you mean here

.github/copilot-instructions.md
bbakernoaa: It's a good thought. Can you provide a suggestion?

.github/copilot-instructions.md
bbakernoaa: In vscode if you create a local conda environment they put it .conda

monetio/readers/amdar.py
bbakernoaa: We are using UGRID and CF convention for everything else and the downstream repos for regridding and such natively understand these. I wasn't aware of the NCAR AIRCRAFT format.

monetio/readers/goes.py
bbakernoaa: I don't have a mypy check or something like it at this point.

monetio/readers/drivers.py
bbakernoaa: Yes that's the idea specifically for data that is typically tabular

bbakernoaa:
Thanks @DWesl

Yes this is a massive PR with a large accumulation of features and a completely new structure and API. V1 is to move away from the eager loading and into a lazy loading paradigm. It also is trying to unify readers by having common tools through the class inheritance.

.github/instructions/python.instructions.md
DWesl: For history:
Should this also try to track reductions (or coarsening) along dimensions through [the CF `cell_methods` attribute](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#cell-methods)?
Should this update [the ACDD `date_modified`/`date_metadata_modified` attributes](https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3#Suggested) as appropriate, or track that solely through the `history` attribute?

... I'm realizing that history is only an example, so it should generalize, though it might not hurt to suggest CF (and/or ACDD) as conventions that specify metadata that might need to be updated.

For dropping variables:
Should this also track CF [`ancillary_variables`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#ancillary-data), [`bounds`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#cell-boundaries), [`cell_measures`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#cell-measures), [`grid_mapping`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#grid-mappings-and-projections), and [`formula_terms`](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#parametric-vertical-coordinate)?  

Looking at that list, aware that [`UGRID`](https://ugrid-conventions.github.io/ugrid-conventions/) has its own list of related variables that need to come along, I suspect generalizing the language some would be the better approach: "don't drop coordinate variables or those with information relevant to interpretation, such as those needed by UGRID" might be a decent start.  Doing things through XArray will pull along `grid_mapping` and some of the mesh topology variables automatically, but the `bounds` and some of the other mesh topology variables might get dropped.

.github/copilot-instructions.md
DWesl:
```suggestion
* Prefer imperative language (`must`, `must not`, `never`, `always`) for enforceable rules.  Use definitions from [RFC 2119](https://www.ietf.org/rfc/rfc2119.html).
```
It might creep to using similar language for optional information, but that shouldn't be a problem.  It uppercasing those words might require sticking a link somewhere.

The other option would be:
```suggestion
* Prefer imperative language (`must`, `must not`, `never`, `always`) for enforceable rules.
* Use language from [RFC 2119](https://www.ietf.org/rfc/rfc2119.html) where appropriate for clarity.
```

monetio/readers/amdar.py
DWesl: Hm.  I suppose consistency is as good a reason as any, and UGRID starts off with [1D network topology](https://ugrid-conventions.github.io/ugrid-conventions/#1d-network-topology), so it's applicable, I'm just used to seeing it for triangular- or polygonal-mesh atmospheric models, which airplanes aren't.  You could layer on the other metadata as well; I'll likely wind up doing that with [geozarr](https://geozarr.org/conventions.html) metadata once I work out how to get that working.

monetio/readers/drivers.py
DWesl: Or if you're decently sure nothing else will be modifying `xr_kwargs` in another thread, an alternate might be faster before [Python 3.11](https://docs.python.org/3/whatsnew/3.11.html#misc):
```python
for key in iterools.chain(mfdataset_keys, dataset_driver_keys, ...):
    if key in xr_kwargs:
        del xr_kwargs[key]
```
or
```python
for key in xr_kwargs.keys() & (set(mfdataset_keys) | set(dataset_driver_keys) | ...):
    del xr_kwargs[key]
```
which can be further simplified by defining `mfdataset_keys`, `dataset_driver_keys`, and so forth as [`frozenset`](https://docs.python.org/3/library/stdtypes.html#frozenset) and defining those and their union at module level.

monetio/readers/goes.py
DWesl: It looks like that's more a feature of docstring checkers, which usually do string comparisons for simplicity

monetio/readers/drivers.py
```diff
@@ -75,46 +132,154 @@ def _open_via_icechunk(vds, icechunk_repo: str, virtualizarr_file: str | None) -
...
+        if isinstance(var.data, ManifestArray):
+            for path in var.data.manifest.iter_nonempty_paths():
+                if path.startswith("s3://"):
+                    parts = path[5:].split("/", 1)
+                    prefix = "s3://" + parts[0] + "/"
+                    unique_prefixes.add(prefix)
+                elif path.startswith("http://") or path.startswith("https://"):
+                    parts = path.split("/", 4)
+                    prefix = "/".join(parts[:3]) + "/"
+                    unique_prefixes.add(prefix)
```
DWesl, line 162: I think this would be equivalent:
```suggestion
                from urllib.parse import urlsplit
                url_parts = urlsplit(path)
                if url_parts.scheme in ("s3", "http", "https"):
                    unique_prefixes.add(f"{url_parts.scheme:s}://{url_parts.netloc:s}")
```
of course, you could have `unique_prefixes` be a set of `(scheme, netloc)` tuples, and turn those back into a string when assembling `url_prefix` for the `icechunk.VirtualChunkContainer` below.

monetio/readers/drivers.py
```diff
@@ -75,46 +132,154 @@ def _open_via_icechunk(vds, icechunk_repo: str, virtualizarr_file: str | None) -
...
         elif path.startswith("http://") or path.startswith("https://"):
-            return fsspec.filesystem("http")
+            return fsspec.filesystem("http", **kwargs)
         elif path.startswith("ftp://"):
-            return fsspec.filesystem("ftp")
-        return fsspec.filesystem("file")
+            return fsspec.filesystem("ftp", **kwargs)
+        elif path.startswith("file://"):
+            return fsspec.filesystem("file", **kwargs)
+        return fsspec.filesystem("file", **kwargs)
```
DWesl, line 279: Not shorter, and probably not faster either, but possibly clearer and easier to extend?
```suggestion
        from urllib.parse import urlsplit
        scheme, *_ = urlsplit(path)
        if scheme == "s3":
            # Default to anonymous access for S3 if not specified
            if "anon" not in kwargs and "storage_options" not in kwargs:
                kwargs.update(get_default_storage_options(path))

            return fsspec.filesystem("s3", **kwargs)
        elif scheme in ("http", "https"):
            return fsspec.filesystem("http", **kwargs)
        elif scheme in ("ftp", "file"):
            return fsspec.filesystem(scheme, **kwargs)
        return fsspec.filesystem("file", **kwargs)
```

monetio/readers/drivers.py
```diff
@@ -142,12 +307,24 @@ def expand_paths(path_input: str | list[str], fs=None) -> list[str]:
...
+                # Manual fix for common cases in monetio:
+                protocol = ""
+                possible_protocol, _host_path = path_input.split("://", 1)
+                if possible_protocol in ("s3", "http", "https", "ftp"):
+                    protocol = f"{possible_protocol}://"
+
+                if protocol and not str(files[0]).startswith(protocol):
+                    files = [
+                        f"{protocol}{f}" if not str(f).startswith(protocol) else f for f in files
+                    ]
```
DWesl, line 326: If `protocol` doesn't get set anywhere else, you can condense this a bit:
```suggestion
                    if not str(files[0]).startswith(protocol):
                        files = [
                            f"{protocol}{f}" if not str(f).startswith(protocol) else f for f in files
                        ]
```

PR #282, opened 2026-05-27
Add CESM-SE regridding support with xregrid

PeizhiHao71:
## Summary

This PR updates the CESM-SE reader workflow to improve compatibility with MELODIES-MONET and support processing of unstructured-grid CESM-SE outputs.

### 1. Support regridding unstructured CESM-SE data to regular lat/lon grids

Added support for regridding unstructured CESM-SE model outputs to standard latitude–longitude grids using `xregrid`. This enables downstream workflows that expect structured gridded data.

### 2. Support customizable horizontal grid resolution

The target latitude–longitude grid resolution can now be customized by the user, allowing flexible output resolutions depending on analysis requirements and computational cost.

### 3. Conditional regridding based on `surface_only`

Implemented logic to determine whether regridding is necessary based on the `surface_only` option:

* If `surface_only=True`, the original unstructured CESM-SE grid is preserved, since surface station pairing does not require a regular lat/lon grid.
* If `surface_only=False`, the data are regridded to a standard structured grid for satellite and gridded analysis workflows.

This approach minimizes unnecessary computational cost while preserving compatibility with existing workflows.

### 4. Vertical variable renaming following CESM-FV conventions

Applied the same vertical-variable renaming workflow used in the CESM-FV reader to maintain consistency across CESM model interfaces. This includes renaming and harmonizing multiple vertically resolved variables.

### 5. Standardized model output format for MELODIES-MONET

The updated reader now outputs model datasets in a more unified and standardized format compatible with downstream MELODIES-MONET processing and pairing routines.

## Motivation

The main motivation for this update is to improve support for CESM-SE model outputs within the existing MELODIES-MONET workflow while minimizing changes to the broader code structure and maintaining consistency with existing CESM-FV processing logic.

## Notes

* Regridding functionality relies on `xregrid`.
* The workflow was designed to preserve backward compatibility as much as possible.
* This update primarily targets satellite and gridded comparison workflows while retaining efficient handling for surface-only pairing cases.

### DIFF
--- a/monetio/models/_cesm_se_mm.py
+++ b/monetio/models/_cesm_se_mm.py
@@ -1,7 +1,10 @@
 """CESM File Reader"""
 
 import xarray as xr
-
+import warnings
+import numpy as np
+from xregrid import Regridder, create_global_grid
+from numpy import meshgrid
 
 def open_mfdataset(
     fname,
@@ -9,6 +12,11 @@ def open_mfdataset(
     convert_to_ppb=True,
     var_list=["O3", "NO", "NO2", "lat", "lon"],
     scrip_file="",
+    mesh_file="",
+    regrid_method ='nearest_s2d',
+    lat_resolution = 0.1,
+    lon_resolution = 0.1,
+    surf_only=False,
     **kwargs,
 ):
     """Method to open multiple (or single) CESM SE netcdf files.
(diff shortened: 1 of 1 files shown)


bbakernoaa:
@PeizhiHao71 I don't think the regridding step should occur within the IO piece. It should simply hand back a UGRID convention dataset/array

PR #287, opened 2026-07-28
Fix kolmogorov_zurbenko_filter calling removed pd.rolling_mean

arpitjain099:
`kolmogorov_zurbenko_filter` in `monetio/util.py` calls `pd.rolling_mean`, which pandas deprecated in 0.18 and removed in 0.23. On any pandas that installs alongside a supported Python it blows up on the first iteration:

```
AttributeError: module 'pandas' has no attribute 'rolling_mean'
```

I hit it on pandas 2.3.3 and again on 3.0.5, so the helper cannot run as written.

Switched to `z.rolling(window=window, min_periods=1, center=True).mean()`, the documented replacement, which keeps the same `min_periods` and `center` behaviour. That leaves the function-local `import pandas as pd` unused, so I dropped it; a side effect is that the string underneath it becomes a real docstring instead of a statement that gets thrown away.

Added a test to `tests/test_util.py` that runs the filter over a short series with a spike and checks the spike is smoothed. It fails with the AttributeError before the change and passes after (3 passed locally).

One thing to flag: nothing inside monetio calls this helper, so if you would rather drop it than fix it, that is a fair call and I will close this.


zmoon:
> One thing to flag: nothing inside monetio calls this helper, so if you would rather drop it than fix it, that is a fair call and I will close this.

Thanks for pointing this out, but we might as well fix it. Thanks for also adding a test.

